PR #7, opened 2014-11-11
Initialize m_buffer_out

tempoz:
The constructors for transform_width were not initializing m_buffer_out. This was uncovered by Coverity as issues CID60662 and CID54443.

### DIFF
--- a/include/boost/archive/iterators/transform_width.hpp
+++ b/include/boost/archive/iterators/transform_width.hpp
@@ -113,6 +113,7 @@ public:
     transform_width(BOOST_PFTO_WRAPPER(T) start) : 
         super_t(Base(BOOST_MAKE_PFTO_WRAPPER(static_cast< T >(start)))),
         m_buffer_out_full(false),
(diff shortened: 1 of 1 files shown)


pepsiman:
Default initialization of a character does nothing, but line 116 is performing value initialization which sets it to zero.

tempoz:
Added commit to initialize character to zero; thank you for your help with this issue!

pepsiman:
I mean, the empty brackets are "value initialization", and set chars (or ints, pointers, etc) to zero.

tempoz:
Yeah, sorry pepsiman, I misread your comment and didn't remember the exact rules on value initialization. Upon re-reading them, I am fairly sure the more flexible solution (that will work regardless of the type of m_buffer_out) are the empty brackets, so i will revert the last commit. Thanks!

For any curious, the reference I am using on value initialization can be found here: http://en.cppreference.com/w/cpp/language/value_initialization

PR #8, opened 2015-02-09
Fix warning in variant.hpp

apolukhin:
Fix for the following warning:

D:/Boost/boost/serialization/variant.hpp:69:75: warning: typedef 'types' locally defined but not used [-Wunused-local-typedefs]
     typedef typename boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)>::types types;


apolukhin:
Looks like the issues was fixed

PR #9, opened 2015-03-04
Update register_archive.hpp

akumta:
Per boost ticket# 11004


akumta:
Thank you. 



----- Original Message -----

akumta:
The fix doesn't seem to have made it in to Boost 1.58 beta build RC2.
Any idea why or what needed to have happened to  have got this in?

Is/Was there something that I needed to have done at my end?

Thanks.

PR #19, opened 2015-07-31
Changes to fix serialization for gcc and clang on Windows

eldiener:
These are changes which will fix the building of the serialization library for gcc and clang on Windows. There are still problems building the wide character serialization library for gcc and clang on Windows, but this problem appears to be one of the 'ld' linker involved and so far the mingw(-64) people are not interested in fixing this problem so I am not going to bother trying to get the building of the wide character serialization library on Windows to work properly until they fix 'ld' on Windows.

### DIFF
--- a/include/boost/archive/archive_exception.hpp
+++ b/include/boost/archive/archive_exception.hpp
@@ -87,6 +87,7 @@ public:
         const char * e1 = NULL,
         const char * e2 = NULL
     ) BOOST_NOEXCEPT;
+    BOOST_ARCHIVE_DECL archive_exception(archive_exception const &) BOOST_NOEXCEPT ;
     virtual BOOST_ARCHIVE_DECL ~archive_exception() BOOST_NOEXCEPT_OR_NOTHROW ;
     virtual BOOST_ARCHIVE_DECL const char * what() const BOOST_NOEXCEPT_OR_NOTHROW ;
 };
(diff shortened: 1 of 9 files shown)


eldiener:
I was able to fix the wide character build of the serialization library for mingw-64/gcc/ld on Windows by providing a copy constructor and a destructor for the xml_archive_exception class. The copy constructor is needed because functionality in the wide character serialization library throws objects of xml_archive_exception by value and the destructor, which does nothing, is needed because gcc needs every implicitly defined class member functiion to be explicitly exported/imported if the class is being created and destroyed outside the shared library in which it is defined.

eldiener:
Fixes in the clang trunk for clang 3.8 allows the build of the serialization library to work for clang on Windows targeting mingw(-64)/gcc. Unfortunately clang versions 3.4.1, 3.5.2, and 3.6.2, all on Windows, will not build due to the bug which was only recently fixed on the trunk. It is highly doubtful that this bug fix will be backported to earlier versions. The clang 3.7 version has been closed but has not yet been officially released, but again I do not expect a backport of the recent fix. This is a problem of course with these releases on Windows, not on Linux.

For clang 3.6.2 and 3.5.2 on Windows the normal serialization library will build but the wide character serialization library will not build. For clang 3.4.1 on Windows nothing builds with numerous errors, and I believe the export/import of individual member functions is just broken for clang 3.4.1 on Windows and will not work.

xantares:
:+1: it does fix this 1.59 regression with mingw64/gcc !

the only other lib failling with mingw now is context(+coroutine*) after the 1.58 regression (https://github.com/boostorg/context/issues/16)

eldiener:
I see that the latest 'develop' branch for serialization now works with mingw/gcc so i will close this PR.

PR #22, opened 2015-08-27
Don't use thread unsafe wide character processing functions.

EdSchouten:
The mbtowc(), wctomb() and mblen() functions are not supposed to be
thread-safe, as they have to keep parsing state in a global variable.
The mbrtowc(), wcrtomb() and mbrlen() functions should be used instead.

This change also simplifies how wchar_from_mb<Base>::drain() works.
There is no need to first iterate over the input using mblen().
mbrtowc() can store the partially parsed multibyte character in its
mbstate_t.

### DIFF
--- a/include/boost/archive/impl/xml_iarchive_impl.ipp
+++ b/include/boost/archive/impl/xml_iarchive_impl.ipp
@@ -18,10 +18,11 @@ namespace std{
 #endif
 
 #ifndef BOOST_NO_CWCHAR
-#include <cstdlib> // mbtowc
+#include <cwchar> // mbstate_t and mbrtowc
 #if defined(BOOST_NO_STDC_NAMESPACE)
 namespace std{ 
-    using ::mbtowc;
+    using ::mbstate_t;
+    using ::mbrtowc;
  } // namespace std
 #endif
 #endif // BOOST_NO_CWCHAR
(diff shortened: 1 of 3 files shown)


robertramey:
I've got a slight problem with this.

What if BOOST_NO_CWCHAR is defined?  Then what.  Shouldn't there be an else in there to support compilers which don't have a <cwchar> header?

EdSchouten:
Hi Robert,

For `xml_iarchive_impl.ipp` this change actually does the right thing: we used to test against `BOOST_NO_CWCHAR`, but then used the non-reentrant versions provided by `<cstdlib>` instead. It now actually uses the `<cwchar>` functions.

For the other files the question that I have is: are there actually any operating systems out there that support `mbtowc()`, but not `mbrtowc()` (and that Boost also wants to support)? If so, we could add back the old code and test against `BOOST_NO_CWCHAR` if you want.

robertramey:
"For the other files the question that I have is: are there actually any operating systems out there that support mbtowc(), but not mbrtowc() (and that Boost also wants to support)?"

The short answer is: I have no idea.  Bitter experience has taught me not to second guess what other players, (compiler developers, standard library writers, and users) are going to do.  It's actually easier just to include the wider compatibility.  It might not seem so, but almost all configurations have quirks of some sort so this is the only strategy that really works.  When one runs across these situations, it's a huge amount of time for everyone involved to research the circumstances then make a fix.

 If so, we could add back the old code and test against BOOST_NO_CWCHAR if you want."

I think that would be the best, I would appreciate your doing this.

Thanks for you help.

Robert Ramey

robertramey:
Hmm - I merged this in, pulled in the changes and run the tests and all the xml_warchive tests failed.  Did you actually test this?  I'm using Clang on Mac OS X

PR #26, opened 2015-09-18
Fix compilation

MarcelRaad:
BOOST_STATIC_ASSERT_UNUSED_ATTRIBUTE was removed in https://github.com/boostorg/static_assert/commit/87f3be4a8fe6fa32646aa01782dda29e6eeb0b4c and replaced by BOOST_ATTRIBUTE_UNUSED.


MarcelRaad:
Seems like this is now fixed in develop. Thanks!

PR #30, opened 2015-10-17
Change archives to backup and restore the C++ locale:

jzmaddock:
avoids double delete of the locale's facets.
Move utf8_codecvt_facet.cpp into main library as it's referenced from there.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -80,6 +80,7 @@ SOURCES =
     xml_iarchive
     xml_oarchive
     xml_archive_exception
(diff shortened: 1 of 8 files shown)


jzmaddock:
Note: this PR does not completely fix the msvc tests - there are unresolved externals when building the dll version of the library.  However, building with link=static passes most tests now (a couple of those also fail with unresolved externals).

robertramey:
I think this is fixed.  There are some changes susequent to these so it's hard to check

PR #33, opened 2015-11-01
Fix MSVC compilation errors

MarcelRaad:
BOOST_SERIALIZATION_SOURCE has to be defined before the first
serialization/config.hpp include. As singleton.hpp also includes config.hpp
now, BOOST_SERIALIZATION_DECL was defined as BOOST_SYMBOL_IMPORT instead of
BOOST_SYMBOL_EXPORT in the serialization source files, leading to C2491
"definition of dllimport function not allowed" errors on MSVC when building
serialization as a shared library.


MarcelRaad:
Obsolete.

PR #36, opened 2016-02-10
Add missing includes for some stl containers

kartikmohta:
Just including the corresponding header for the stl container should work, but leads to a build failure for some of the containers due to missing includes.


kartikmohta:
Without these includes, trivial programs such as
```cpp
#include "boost/serialization/queue.hpp"

int main(void)
{
  return 0;
}
```
fail to build.

kartikmohta:
Looks like these changes have been made in 42b5b6859b757cc0ce3a3e2bef37cc13dea925fc and 8f7b3be1d2104cdc6b5778653cf98ca4dc06a1e6, so closing.

PR #37, opened 2016-02-12
Fix linkage mismatching on MSVC.

Flast:
It breaks windows build.


Flast:
7ae4087c05fd55a7eb9098534c696eca1989b4c4 already focused this issue, close.

PR #40, opened 2016-05-04
`noexcept` specifications for `BOOST_STRONG_TYPEDEF`

praetorian20:
`BOOST_STRONG_TYPEDEF` now detects whether the type being wrapped is
nothrow default/copy constructible and nothrow assignable. The wrapped
type is given its own `noexcept` specifications accordingly.

PR #43, opened 2016-05-31
Disable `BOOST_STRONG_TYPEDEF` tests for pre-C++11 compilers

praetorian20:
The previous commit added `noexcept` specification for
`BOOST_STRONG_TYPEDEF`, but the unit test added didn't consider
pre-C++11 compilers where `noexcept` is not available. The
`static_assert`s in the unit test have now been disabled for these
compilers.


praetorian20:
This should address the test failures in #40. I just disabled all the `static_assert`s in the unit tests for pre-C++11 compilers.

PR #44, opened 2016-07-05
Explicit narrowing conversion

thtrummer:
http://www.boost.org/development/tests/develop/developer/output/timber-cygwin-boost-bin-v2-libs-serialization-test-test_codecvt_null-test-gcc-5-4-0-debug.html

### DIFF
--- a/test/test_codecvt_null.cpp
+++ b/test/test_codecvt_null.cpp
@@ -40,29 +40,29 @@ struct test_data
 
 template<>
 wchar_t test_data<2>::wchar_encoding[] = {
-    0x0001,
-    0x007f,
-    0x0080,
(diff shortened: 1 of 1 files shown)


robertramey:
are you able to get gcc tests to actually work on serialization develop?  The work on my machine but they all fail on the test matrix.  I realize that this is a separate issue but I'm desperate here.

thtrummer:
Not really. There are some compiler errors but most of the tests just core dump. This is probably caused by a problem in Boost.Filesystem so I'm currently trying to get these tests working.

robertramey:
OK - I'm rolling this in from my end - feel free to close when you've checked it out

PR #45, opened 2016-07-10
Remove extra ";"s from strong type members

aw1621107:
Yes, this is incredibly trivial. Submitted because it popped up in warnings and was an easy fix.

Feel free to close this PR and include the contents in another more substantial one.


robertramey:
OK - it's a trivial fix - but at least I can accomplish SOMETHING today.  I'll roll it in from my end.

PR #50, opened 2017-01-17
constness fix in (make_)binary_object

el-bart:


el-bart:
please double-check this PR - i had issues with official build, and finally build the test manually...

robertramey:
What the the motivation for this.  Wasn't test_binary.cpp complete enough?

el-bart:
i've spotted the const-correctness issue, while serializing collection. test code example did not compile, until consts got uncommented ("save()" is const). there is a const-cast under the hood any way, but now it does not leak to user's code.

PR #51, opened 2017-01-21
Build all source for library with BOOST_SERIALIZATION_SOURCE solves s…

eldiener:
…ingleton link problem. I have placed the BOOST_SERIALIZATION_SOURCE definition in the build jamfile so it applies to all the non-wide source, and removed it from each source file. Alternatively, if desired I could remove it from the jamfile and add it to each source file, but I felt the way I did it was easier. The important point to solving the singleton link problem for gcc under Windows is that the definition must apply to each source file in the build so that the build of the non-wide character serialization library sees the necessary classes/functions as exported.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -107,6 +107,7 @@ lib boost_serialization
     <toolset>clang:<cxxflags>"-ftemplate-depth-255"
     <toolset>darwin:<cxxflags>"-ftemplate-depth-255"
     <link>shared:<define>BOOST_SERIALIZATION_DYN_LINK=1
+    <link>shared:<define>BOOST_SERIALIZATION_SOURCE
     ;
 
 lib boost_wserialization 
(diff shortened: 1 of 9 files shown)


robertramey:
"Alternatively, if desired I could remove it from the jamfile and add it to each source file,"

How would that be different than the way it was originally?

eldiener:
Quoting:  

"Alternatively, if desired I could remove it from the jamfile and add it to each source file,"

How would that be different than the way it was originally?

Endquoting:

It would not be different from my PR, except in following your current code to have 
#define BOOST_SERIALIZATION_SOURCE in the source files themselves.

jeking3:
It makes sense to have this definition in the build system and not in the source files.  The headers should key off this and do the right thing for both cases.  Shouldn't the archive definition also be handled the same way?

robertramey:
>It makes sense to have this definition in the build system and not in the source files. The headers should key off this and do the right thing for both cases. Shouldn't the archive definition also be handled the same way?

I think this would conflict with the way that "visibility" works.  It's actually a nightmare as there is a lot of variation among compilers.  I don't think a "sledge hammer" approach is appropriate.

eldiener:
I am glad you finally got serialization working with mingw(-64)/gcc on Windows. Congratulations !

PR #52, opened 2017-02-22
Undefined behaviour in basic_iarchive.cpp

olologin:
I found this UB while using boost serialization at work, and of course, I have to say, it was caused by incorrect usage of boost::serialize from my side, because I tried to deserialize object of type, that wasn't actually in container. Anyway, I don't think this should be the cause for undefined behaviour. It should throw at least some exception in release build. The problem was detected accidentally in debug build, "subscript out of range" exception was thrown by vector [] operator.

Sorry that I don't have reproducible example of serialization/deserialization that causes the problem. I'm too lazy to simplify our closed code. I hope people that are good at this great library will understand cause of this problem without examples.


robertramey:
this looks like a side effect from some other error.  How did this happen in the first place?

PR #54, opened 2017-06-01
Included boost/throw_exception.hpp when exceptions are disabled.

dgrafov:
There is a compilation error when boost is compiled without exceptions: boost::serialization::throw_exception calls boost::throw_exception which is not declared in this case. The declaration is situated in boost/throw_exception.hpp.

### DIFF
--- a/include/boost/serialization/throw_exception.hpp
+++ b/include/boost/serialization/throw_exception.hpp
@@ -19,6 +19,8 @@
 
 #ifndef BOOST_NO_EXCEPTIONS
 #include <exception>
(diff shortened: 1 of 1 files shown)


dgrafov:
@robertramey it gives a compiler error when I use binary_from_base64 without exception and compiler points to this file, so it looked reasonable to me, that if boost::throw_exception is called, the file with it's declaration should be included. But now I see, that there is no include by intent and I can fix the compilation error in my own code. So I'm closing my pull-request.

PR #55, opened 2017-06-10
Avoid undefined behavior.

jsonn:
Macros expanding to defined() expressions are undefined, move the conditional logic to outside the expression.


jsonn:
This is the same as ticket 12144 and 8120. Looking at the latter, the explicit MSVC can most likely be dropped.

masterleinad:
This is the same as  the recent commit 8d0b0db5deb54fd091e8467364b5ec293a9a8c6b.

PR #65, opened 2017-09-28
Use mingw

jschueller:
It seems some work has begun to use mingw on appveyor, this goes a little further so it compiles and builds some tests but not all succeed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -56,12 +56,12 @@ install:
   - git submodule init tools/build
   - git submodule update
   - xcopy /s /e /q %APPVEYOR_BUILD_FOLDER% libs\serialization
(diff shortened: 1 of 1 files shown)


jschueller:
see #63

robertramey:
Hmmm - How come this was closed?  Should it not be merged in? Looks like it made things work.  and I had a couple of questions:

..\..\..\b2 target-os=windows toolset=gcc variant=release threading=multi link=static,shared runtime-link=shared address-model=32 architecture=x86 cxxflags="-std=c++11

 -O2 // why use this with varian=release

-g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions --param=ssp-buffer-size=4" // what is all this stuff?

jschueller:
It looked like #63 is duplicate.
I just copied the standard flags when compiling mingw packages on Fedora.

PR #66, opened 2017-10-06
Clang warning cleanup on preprocessor

drivehappy:
This should resolve this ticket: https://svn.boost.org/trac10/ticket/12144 where clang is reporting a warning on "-Wexpansion-to-defined"


drivehappy:
Nvm, looks like it was just addressed yesterday with cd2f77d

PR #67, opened 2017-10-09
Cleanup: Silence Xcode 8.2.1 warnings.

KevinHopps:


robertramey:
normally I don't address warning since they vary from compiler to compiler and it too easily turns into a game of whack-a-mole.  But in this case I'll make an exception

KevinHopps:
Thank you. These are errors for Photoshop, since we treat warnings as errors. I re-make these changes each time I download the latest boost. This is the first time I have pushed some of my changes back.

PR #68, opened 2017-10-12
fix uninitialized initialized flag

jeking3:
https://scan4.coverity.com/reports.htm#v17121/p12449/fileInstanceId=26953615&defectInstanceId=5367075&mergedDefectId=192130


robertramey:
merged in

PR #83, opened 2017-12-24
Fix narrowing.

DanielaE:
Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/archive/iterators/unescape.hpp
+++ b/include/boost/archive/iterators/unescape.hpp
@@ -54,7 +54,7 @@ public:
 private:
     value_type dereference_impl() {
         if(! m_full){
-            m_current_value = static_cast<Derived *>(this)->drain();
+            m_current_value = static_cast<value_type>(static_cast<Derived *>(this)->drain());
             m_full = true;
         }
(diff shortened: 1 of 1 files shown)


robertramey:
The file unescape.hpp contains the text:

// class used by text archives to translate char strings to wchar_t
// strings of the currently selected locale

which suggests to me that the value type should wide type in any case.  Also look through the code I can't find where the Derived type wouldn't be a wide type.  Could you tell which tests emits this error, the compiler and switches used to reproduce it?

DanielaE:
`test_iterators` exhibits this narrowing issue with every version of MSVC as the Boost test matrix clearly shows. The other compilers don't, but just by pure luck because they happen to compile without narrowing warnings enabled.

robertramey:
OK - found it at xml_unescape line # 45.  This was part of a workaround for msvc 7.1.  Since msvc 7.1 is no longer supported (since about 10 years ago!).  I just eliminated that work around.  This should address the issue.

robertramey:
Damn - turns out that the ICE crops up for all microsoft compilers of versions < 14

PR #84, opened 2017-12-24
replace uncaught_exception with uncaught_exceptions

DanielaE:
The former is deprecated in C++17 and replaced by the latter.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/archive/impl/basic_text_oprimitive.ipp
+++ b/include/boost/archive/impl/basic_text_oprimitive.ipp
@@ -8,6 +8,12 @@
 
 //  See http://www.boost.org for updates, documentation, and revision history.
 
+#if __cplusplus > 201402 || (defined(_MSVC_LANG) && _MSVC_LANG > 201402)
+# define BOOST_ARCHIVE_UNCAUGHT_EXCEPTION() std::uncaught_exceptions()
+#else
+# define BOOST_ARCHIVE_UNCAUGHT_EXCEPTION() std::uncaught_exception()
+#endif
+
 #include <cstddef> // NULL
 #include <algorithm> // std::copy
 #include <exception> // std::uncaught_exception
@@ -106,7 +112,7 @@ basic_text_oprimitive<OStream>::basic_text_oprimitive(
 template<class OStream>
 BOOST_ARCHIVE_OR_WARCHIVE_DECL
 basic_text_oprimitive<OStream>::~basic_text_oprimitive(){
-    if(std::uncaught_exception())
+    if(BOOST_ARCHIVE_UNCAUGHT_EXCEPTION())
         return;
     os << std::endl;
 }
--- a/include/boost/archive/impl/xml_iarchive_impl.ipp
+++ b/include/boost/archive/impl/xml_iarchive_impl.ipp
@@ -8,6 +8,12 @@
 
 //  See http://www.boost.org for updates, documentation, and revision history.
 
+#if __cplusplus > 201402 || (defined(_MSVC_LANG) && _MSVC_LANG > 201402)
+# define BOOST_ARCHIVE_UNCAUGHT_EXCEPTION() std::uncaught_exceptions()
+#else
+# define BOOST_ARCHIVE_UNCAUGHT_EXCEPTION() std::uncaught_exception()
+#endif
+
 #include <boost/config.hpp>
 #include <cstring> // memcpy
 #include <cstddef> // NULL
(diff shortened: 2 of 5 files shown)


robertramey:
Consider getting macros for all C++ 17 features added to boost/config.  Then make make conditions which depend on these macros.  This has turned out to be a much more permanent and scalable manner of keeping the serialization library useable on C++03,...,C++17,...   Thanks for your efforts.  I depend a lot on help from my friends to keep this  this thing relevant.

robertramey:
I've looked at this a little bit more.  I think it's a bigger issue than changing from to uncaught_exceptions.  I never really considered how exceptional conditions should be handled when detected from within a destructor.  I know it's been discussed a lot, but following these discussions makes my head hurt.  I'm thinking maybe terminate() should be used but I don't really know.  For the moment, I'm not going to mess with this.  If you want to investigate this is more detail, I would appreciate it.

pavelkryukov:
Just to draw your attention: GCC 8 warns about deprecating `std::uncaught_exception()`, generating errors in `-Werror` compilation mode.

jeking3:
Probably a good idea to move common stuff down into Boost.Config so everyone can benefit.

DanielaE:
James, I agree with moving this stuff to Boost.Config rather than duplicating it all over or implementing it in different ways. But I don't agree with making things needlessly complicated and reaching out to undocumented compiler-specifics when a simple and portable solution will do the job. Well at least until Robert has got a clue what to do with the information from `uncaught_exception/s` in the first place, lest even the number of uncaught ones.

jeking3:
I'm still weary of using `__cplusplus` because of [MSVC's behavior](https://developercommunity.visualstudio.com/content/problem/139261/msvc-incorrectly-defines-cplusplus.html) there.  That's why I believe this area needs more investigation and a better implementation.

DanielaE:
Therefore the check for MSVC_LANG.
It works, we're a MSVC-only company here, using a wide variety of MSVC versions and modes.

jeking3:
Very well, however we should still move this to Boost.Config and have all three implementations use it.  Nevertheless, this could be done after.

jeking3:
@pdimov any thoughts on a good Boost.Config macro name for this?  `BOOST_STD_UNCAUGHT_EXCEPTIONS` perhaps?

pdimov:
The convention would call for `BOOST_NO_CXX14_UNCAUGHT_EXCEPTIONS`. (https://www.boost.org/doc/libs/1_68_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.macros_that_describe_c__14_features_not_supported)

pdimov:
Or rather, `BOOST_NO_CXX17_UNCAUGHT_EXCEPTIONS`, as this is in C++17, not in C++14.

jeking3:
So you wouldn't recommend defining a macro that can be used as `BOOST_STD_UNCAUGHT_EXCEPTIONS()` but instead forcing the implementer to have an #if #else #endif in every case?

pdimov:
It's not possible to have `BOOST_STD_UNCAUGHT_EXCEPTIONS()`. You could have `BOOST_STD_UNCAUGHT_EXCEPTION()` (without the S) because the old function is implementable in terms of the new one, but not vice versa.

But using such a macro kind of misses the point; the new function was added because the old function doesn't quite work. The correct use is to record the number of active exceptions in the constructor, and then check in the destructor if the new number of active exceptions is higher.

It works as a minimal fix for old code as in this example, but it shouldn't be what new code would use.

pdimov:
Or, reading back this thread, perhaps @Lastique could move his implementation of `uncaught_exceptions` from Log to Core?

Lastique:
I'm ok with moving it for public use, but maybe Boost.Exception is a better place? I would cc Emil here, but I don't know his username.

Lastique:
@zajo What do you think if we move `std::uncaught_exceptions` implementation from Boost.Log to Boost.Exception? See https://github.com/boostorg/log/blob/develop/src/unhandled_exception_count.cpp.

pdimov:
No, Core is better.

Lastique:
@pdimov Why?

pdimov:
That's where we put standard functions such as f.ex. `std::quick_exit` that are missing in earlier standards. Arguably we should have a dedicated library for that but we don't.

pdimov:
(`std::exchange` too.)

Lastique:
Yes, but Boost.Exception also implements `exception_ptr`, `current_exception`, `rethrow_exception` and ultimately a library about exceptions.

Lastique:
Ok, I've extracted the function as `uncaught_exceptions` to Boost core in https://github.com/boostorg/core/commit/82957de97029e065af787ddc9f4a21e786dfec4a.

robertramey:
So what's the final verdict?  Can I just brainlessly merge this in?  Or am In in for another death march?

Lastique:
My suggestion would be to use `boost::uncaught_exceptions` unconditionally.

Lastique:
`boost::uncaught_exceptions` does not require C++11.

Lastique:
This can be closed because https://github.com/boostorg/serialization/pull/140 was merged.

PR #85, opened 2017-12-27
trac-9882: fix -Wshadow errors in serialization (seen during date_time build)

jeking3:
https://svn.boost.org/trac10/ticket/9882


robertramey:
I've looked at this and I'm not yet convinced.  It seems that this is dependent on the usage of the date-time library and should be fixed there.

jeking3:
I will have to look at it again with your findings in mind.  For now let's leave this open?

PR #87, opened 2018-01-20
Refresh CI environment for more coverage

jeking3:
After adding ubsan builds to Datetime and Uuid, there were some issues in serialization.   This pull request adds a ubsan build job to serialization as an allowed failure so that issues can be uncovered.


robertramey:
I'm OK with this. But it's not passing CI tests.  I'm guessing this is due to unrelated issues.  So for the moment, I'm going to let this sit a little bit until the CI starts to pass with all the other changes.

jeking3:
I will update this PR with a small fix, hopefully tonight; see:
https://github.com/boostorg/format/commit/76293b18b1734c61ff6ada2dc3065ca4d52c5636

robertramey:
I would like to merge this and close it.  But I notice that there are some failures pending. What should I do with this?

jeking3:
On the ubsan build job I just pushed a fix for the issue there.  I still expect it to fail, but I have marked ubsan and valgrind as non-fatal CI errors for you (although I think they should be looked at and there are open issues for them).

The clang dynamic linux job seems like a legitimite issue however, segfault.

jeking3:
@robertramey if you have a chance to log into AppVeyor and cancel builds 191, 192 it will same some time (about 6 hours).

jeking3:
Okay, I'll do it on my fork then.  I'll get back to you with defects identified.

PR #89, opened 2018-01-30
Fix mingw ICE

res2k:


res2k:
Sorry, mistaken submission.

PR #90, opened 2018-01-30
Pull container_hash git submodule on appveyor

res2k:
This makes the tests build again. (Well, except for those producing an ICE on MinGW.)

PR #91, opened 2018-01-30
Fix MinGW ICE

res2k:
Just fiddling with the displacement is apparently enough to avoid triggering the ICE on MinGW.

(Note: Superset of #90)


res2k:
Superseded by #94

PR #96, opened 2018-02-24
Fix object tracking for variant

ricardocosme:
Fix object tracking for a pointer to an object contained into a variant that is an element of a container which loads a `value_type` into a local object. For example a `std::map` with a `boost::variant` as a `mapped_type`.

This commit supports:
- map
- multimap
- unordered_map
- unordered_multimap
- set
- multiset
- unordered_set
- unordered_multiset


ricardocosme:
Build failed for some platforms. I will fix this.

PR #112, opened 2018-07-31
Add missing template args to unordered_map<> serialization

res2k:
Both `std::unordered_map<>` and `boost::unordered_map<>` miss the template argument for the value.
Probably went unnoticed as this would only surface with a non-standard allocator.


res2k:
> This looks like a no brainer. I'm merging into develop now. Subject to subsequent testing and merge to master.

Ping?
Despite your statement I didn't see these changes making it to develop branch yet.

Issue #119, opened 2018-09-24
Null pointers lead to segfault using clang


sgiraudot:
If I understand correctly, Boost Serialization is supposed to handle null pointers internally (handling them with a specific value and not calling `serialize()` on the dereferenced pointer).

I found out that this doesn't work if I compile my program with `clang`.

Consider the following program:

```c++
#include <iostream>
#include <cstdlib>

#include <boost/archive/text_iarchive.hpp>
#include <boost/archive/text_oarchive.hpp>

struct Node
{
  std::string name;
  Node* next;

  Node(const std::string& name)
    : name (name), next(NULL)
  { }

  template <typename Archive>
  void serialize(Archive& ar, unsigned /* version */)
  {
    ar & name;
    ar & next;
  }
};

int main()
{
  Node* first = new Node("first");
  Node* second = new Node("second");
  Node* last = new Node("last");

  first->next = second;
  second->next = last;

  boost::archive::text_oarchive oa { std::cout };
  oa << *first;

  delete first;
  delete second;
  delete last;

  return EXIT_SUCCESS;
};
```

If I compile this with `g++` (version 6.4.0), I get the following output:
```sh
22 serialization::archive 15 1 0
0 5 first 0
1 6 second 0
2 4 last -1
```

Which makes sense. If I compile the same code using `clang` (version 6.0.0), then I get the following output/error:
```sh
22 serialization::archive 15 1 0
0 5 first 0
1 6 second 0
2 4 last 0
zsh: segmentation fault (core dumped)  ./serialize_bug
```

It seems that the NULL pointer is not handled well and that it is dereferenced (leading to the segfault). The bug is still here if I use `nullptr` instead of `NULL`.

The bug disappears for `clang` if I deactivate the optimizations (so I'm not exactly sure if this is a Boost bug or a CLang bug).

robertramey:
OK - good test case.  We'll look into it.

robertramey:
I've rebuilt and run this case with clang under Xcode on my MacOS system.  The output is what I would expect:

`22 serialization::archive 17 1 0
0 5 first 0
1 6 second 0
2 4 last -1
Program ended with exit code: 0`

Note the -1 which is the special tag for null pointers.  Your "correct" output doesn't have this.  Since I can't reproduce it on my own machine it's hard for me to track down

[...]

PR #126, opened 2018-10-06
add unit test and release notes about #119 

pauldreik:
I added a simplified version of the code by @sgiraudot in https://github.com/boostorg/serialization/issues/119#issue-363122998. The unit test triggers null pointer dereference in 1.65 but not latest develop as of now.
Besides the unit test, I added information to the release notes as this is exactly the kind of information at least I expect to see there. I read the release notes to find out if it is relevant to upgrade to a new boost version. Avoiding undefined behaviour in my code is certainly a reason to upgrade!

Edit: I realize I accidentally used nullptr which would break pre C++ 11 builds, is this a problem?


test/test_member_pointer_issue119.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * The code is a minimized and simplifed version of the code example
+ * by Simon Giraudoty, see https://github.com/boostorg/serialization/issues/119
+ * which triggers the error in earlier versions of boost (<1.66), at least
+ * if building with clang 6 with -O3 and definitly if using undefined behaviour
+ * sanitizer.
+ */
+
+struct Node
+{
+  Node* ptr=nullptr;
```
jeking3: Change this so it works on C++03.

robertramey:
Since this issue was resolved over a year ago, I'm not going to add in this tests.  I see you're a first time contributor.  My GitHub foo is not enough to know whether this is first time for this project or first time contributor to GitHub.  But no matter.  I appreciate your efforts and don't want to discourage further contributions in the future.  You've done a nice job here.

pauldreik:
Ok, about the unit test, but what about the additions to the release notes? That is the exact type of information I am looking for, when considering boost upgrades. I thinks it is more important,
the more capable the compilers get taking advantage of undefined behaviour. As stated in #119, it actually happens to real people, it's not hypothetical. Please reconsider incorporating this into the release notes.
Thanks,
Paul

PR #127, opened 2018-10-14
More tests on CI

Flamefire:
This builds feature branches on appveyor as it does for travis -> consistency

It further adds more test runs for release builds, so that is properly tested. This is required as especially with Boost.Serialization optimization may remove classes that are required to stay although they are not referenced. Those visibility macros are in place to allow this, these tests should check them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,20 +27,16 @@ matrix:
   include:
     - os: linux
       compiler: g++
-      env: TOOLSET=gcc VARIANT=debug
+      env: TOOLSET=gcc
 
     - os: linux
       compiler: g++
-      env: TOOLSET=gcc VARIANT=release
-
-    - os: linux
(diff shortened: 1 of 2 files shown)


jeking3:
Splitting out debug and release to their own job isn't efficient.  See my recent merged PRs into assign and logic for .travis.yml; these combine things into one build job per compiler, which is more efficient.  You can use multiple cxxstd and variant in the same build job and this avoids the overhead of environment setup.

Flamefire:
You are right: 90b8ea0f3d0a836b9957b373c4b7154a89e74ac0

I think it would also be a good idea to combine static/shared builds? It was explicitely split probably to have failures show where they occurred on top-level, but again: More overhead.

robertramey:
Resolved ?! - if not open a new PR

PR #128, opened 2018-10-16
Show crash of #111 in a different way

Flamefire:
As @robertramey mentioned that the use of `extended_type_info` in #111 is invalid (can I have the related documentation for this?) I changed the example to be based on the official docu.

Base is the basic serialization in the tutorial: https://www.boost.org/doc/libs/1_67_0/libs/serialization/doc/tutorial.html
Additionally I want a specific level as described here: https://www.boost.org/doc/libs/1_67_0/libs/serialization/doc/traits.html#level

Hence this code should now show the valid use of 2 shared libraries (possibly from different vendors and completely unrelated) each using static Boost.Serialization in a straight forward way.

But it will show the exact same crash as in #111 due to the destruction order mixup with shared libraries on linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ install:
 script:
   - |-
     echo "using $TOOLSET : : $TRAVIS_COMPILER ;" > ~/user-config.jam
-  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared}
+  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared} cxxflags=-fPIC cflags=-fPIC
 
 notifications:
   email:
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -52,6 +52,22 @@ lib dll_polymorphic_derived2
         <link>shared:<define>BOOST_SERIALIZATION_DYN_LINK=1 
     ;
 
+lib multi_shared_1
+    :
+        multi_shared1.cpp
+        ../build//boost_serialization/<link>static
+    :
+        <link>shared
+    ;
+
+lib multi_shared_2
+    :
+        multi_shared2.cpp
+        ../build//boost_serialization/<link>static
+    :
+        <link>shared
+    ;
+
 test-suite "serialization" :
      [ test-bsl-run_files test_array : A : :  [ requires cxx11_hdr_array ] ] # BOOST_NO_CXX11_HDR_ARRAY
      [ test-bsl-run_files test_boost_array : A ]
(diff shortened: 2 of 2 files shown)


robertramey:
Right - I meant without error

Flamefire:
Rebased on current develop. As explained on ML it still crashes like before:

> *** Error in `bin.v2/libs/serialization/test/test_multi_shared_lib.test/gcc-4.8/debug/link-static/visibility-hidden/test_multi_shared_lib': double free or corruption (fasttop):

robertramey:

Flamefire:
@robertramey Neither this nor #111 was included in the current source as far as I could see. I think this test should still be included as it shows the implementation defined behaviour and why the runtime checks instead of asserts are required. Which one would you like? Any changes required?

robertramey:
I had that test included earlier just to see what it would do to the test matrix.  It passed on every platform tested.  I removed it for a couple of reasons.  

a) It didn't fail anywhere so I doubted it's validity and utility.  I wasn't convinced that it actually isolated a problem.  That is, if there's a problem somewhere, I'd like to see a test isolate it.  I think this test needs more work to do what it purports to do.

b) It wasn't to the standard you've set by test_singleton_inherited, etc as far as comments, justification, explanation, copyright etc. A test added to the library becomes part of the library almost forever. From an esthetic viewpoint, I believe that it's valuable that we try to maintain a level of consistency. One could make the case that it's just syntactic sugar and hence irrelevant.  But I think it's helpful to persons in the future to be able to understand the library and how to fix it if necessary.

Soooo - I don't object in principle object to including a test like this.  But I think this one needs more work.

This whole episode does make the idea of separating the singleton into a separate boost library.  I've included features - basically the connected of "locked" and "is_destroyed" to support what the serialization library needs.  Separating this component into a separate library would result in a very interesting discussion as to what such a library should  contain.  I added in these features because I'm paranoid and trust no one.  But maybe that's just me.  Then there is the question of forcing initialization to occur before main.  And the question as to the way I handled the forcing of correct sequencing of dependent static objects.

Making singleton a separate library would make things simpler.  I believe that part of the confusion created in this thread is due to singleton and the serialization and extended type info, etc seeming to be coupled when in fact they are not.  Factoring things in composable components promotes correctness and simplifies testing and maintenance.

Flamefire:
> I had that test included earlier just to see what it would do to the test matrix. It passed on every platform tested

With my fix of the singleton included this is expected. Did you test this **with** my fix or without? If without, did you try on linux? Do you have a link to that test (travis etc.)? Because this DID show the crash (see https://github.com/boostorg/serialization/pull/128#issuecomment-433300799)

As the singleton fix also removed the asserts this can now not show any more failures.

But: This test creates a very specific situation: Using static Boost.Serialization from 2 shared libraries which did crash before and there is no other test like this ATM. These 2 facts make it worth being included.

I'll happily create more docu to point out what this tests and what it does not: It does not test a specific code path but rather how this library handles the implementation defined behaviour.

> I believe that part of the confusion created in this thread is due to singleton and the serialization and extended type info, etc seeming to be coupled when in fact they are not.

`extended type info` uses the singleton. And the way it was used was wrong: There is a case where this assert does not hold: https://github.com/boostorg/serialization/commit/f297d80d6ef55ac66525320d0477c17806aa57cd#diff-d9472ab68b7d3466cc64073e4d16d2ecL98

So this is not really a test of `singleton` but from more of the library: The coupling of the different singletons used.

To get constructive: What I would add/clarify in the test docu:

- Intended to test the combination of static boost in multiple shared libraries
- Reproduces a former bug (at least the triggered asserts) in the coupling of internal components
- This bug came from different destruction order due to implementation defined behaviour of shared libs

What else should be clarified/added?

PR #129, opened 2018-10-17
Add tests for singleton::is_destroyed

Flamefire:
This is basically the same test as in #110.

However in #105 @pdimov noted, that it might be easier, to duplicate some test code and split it into 2 separate test cases (https://github.com/boostorg/serialization/pull/105#discussion_r225581561), so this is done here.

As mentioned in the [docu](https://www.boost.org/doc/libs/1_67_0/libs/serialization/doc/singleton.html) the singleton may be used in 2 cases and "Both are used in the serialization library. ":

- plain: simple class used as singleton<Foo>
- inherited: Foo inherits from singleton<Foo> and can be used directly

Furthermore there is `is_destroyed` which shall "Return true if the destructor on this singleton has been called. Otherwise, return false."

This behaviour is never tested which is the reason for this PR.

Assumption: Classes are destructed in the reverse order of the return from their constructor (see C++ standard) and static function objects are destructed at program termination. This is the base for the `controller` class:

- Accessed from the singleton under test `Foo` in its constructor -> `controller` is destructed after `Foo`
- `Foo`s ctor AND dtor will set a variable in the `controller` to the current state
- This state can be used in the dtor of `controller` to check that the singleton was destroyed (if it wasn't there would be a memory leak)
- The controller can also check `is_destroyed` of the singleton which must be `true` due to the former
- The main function checks the `is_destroyed` returns `false` and accesses any member of `Foo` to avoid its construction being eliminated.

As explained multiple times `is_destroyed` does not return `true` as it should for the plain singleton. #105 contains the fix for this (will rebase #105 when this is merged)

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -149,6 +149,8 @@ if ! $(BOOST_ARCHIVE_LIST) {
         [ test-bsl-run test_smart_cast ]
         [ test-bsl-run test_codecvt_null ]
         [ test-bsl-run test_singleton ]
+        [ test-bsl-run test_singleton_plain ]
+        [ test-bsl-run test_singleton_inherited ]
 
         # [ test-bsl-run test_z ]
 
--- a/test/test_singleton_inherited.cpp
+++ b/test/test_singleton_inherited.cpp
@@ -1,0 +1,82 @@
+/////////1/////////2/////////3/////////4/////////5/////////6/////////7/////////8
+// test_singleton_inherited.cpp:
+// Test the singleton class for a "inherited" singleton (used as Foo:public singleton<Foo>)
+// This can be uses as singleton<Foo>::get_const_instance() OR Foo::get_const_instance()
+//
+// - is_destroyed returns false when singleton is active or uninitialized
+// - is_destroyed returns true when singleton is destructed
+// - the singleton is eventually destructed (no memory leak)
+
+// (C) Copyright 2018 Alexander Grund
+// Use, modification and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include "test_tools.hpp"
+#include <boost/serialization/singleton.hpp>
+#include <boost/preprocessor/stringize.hpp>
+#include <stdexcept>
+
+// Can't use BOOST_TEST because:
+// a) destructors are called after program exit
+// b) This is intended to be used by shared libraries too which would then need their own report_errors call
+// We halso have to disable the Wterminate warning as we call this from dtors
+// C++ will terminate the program in such cases which is OK here
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wterminate"
+#define THROW_ON_FALSE(cond) if(!(cond)) throw std::runtime_error(__FILE__ "(" BOOST_PP_STRINGIZE(__LINE__) ") Assertion failed: " #cond)
+
+// Enum to designate the state of the singletonized instances
+enum ConstructionState{CS_UNINIT, CS_INIT, CS_DESTROYED};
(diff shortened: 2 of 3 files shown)


Flamefire:
Good idea: #131 is that

Flamefire:
Rebased on develop. Still shows the bug as before:

> what():  libs/serialization/test/test_singleton_plain.cpp(66) Assertion failed: boost::serialization::singleton<Foo>::is_destroyed()

jeking3:
The existing CI builds failed:
```
testing.capture-output ..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.run
====== BEGIN OUTPUT ======
No errors detected.
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
terminate called after throwing an instance of 'std::runtime_error'
  what():  test_singleton_plain.cpp(66) Assertion failed: boost::serialization::singleton<Foo>::is_destroyed()
 
EXIT STATUS: 3 
====== END OUTPUT ======
    set Path=C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin;C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\lib;C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\lib32;C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\lib64;%Path%
    set status=0
    if %status% NEQ 0 (
        echo Skipping test execution due to testing.execute=off
        exit 0
    )
     "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.exe"   > "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.output" 2>&1 
    set status=%ERRORLEVEL%
    echo. >> "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.output"
    echo EXIT STATUS: %status% >> "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.output"
    if %status% EQU 0 (
        copy "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.output" "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.run"
    )
    set verbose=0
    if %status% NEQ 0 (
        set verbose=1
    )
    if %verbose% EQU 1 (
        echo ====== BEGIN OUTPUT ======
        type "..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.output"
        echo ====== END OUTPUT ======
    )
    exit %status%
...failed testing.capture-output ..\..\..\bin.v2\libs\serialization\test\test_singleton_plain.test\gcc-6.3.0\debug\link-static\visibility-hidden\test_singleton_plain.run...
```

jeking3:
I think it would be nice to close all the previous pull requests currently outstanding on this issue.  We've got this one showing the issue, and the follow-up.  I'm still concerned about Boost.Serialization and 1.69.0 however, since it looks like Windows builds are broken on develop.

Flamefire:
> The existing CI builds failed:

This is the error I mentioned and wanted to show in this PR

robertramey:

PR #131, opened 2018-10-18
Fix singleton::is_destroyed and crash when using shared libraries

Flamefire:
This supersedes #105 with a more minimal approach based on current develop. Reason: Current develop contains a change (https://github.com/boostorg/serialization/commit/e1893dd28c408bb25352596e2f91892b3d570164), which I deemed mislead and reverted in #105 (5248dee2528b32907003042da84c5f4d650c470e). Note that the PR predates that change!   
To have a smaller changeset this revert is not included here.

Additionally it includes the crash test #111 and the improved is_destroyed test #129 to show that this does indeed fix the problem. Those PRs should be merged first!

The actual fix is https://github.com/boostorg/serialization/commit/c2c6e63c3e72a731e3a64afd8c3549b08f69344f which does the following:  

When a plain singleton `Foo` is used ("plain"=="does not inherit from `singleton<Foo>`) the class `singleton<Foo>` is never instantiated. Hence it is never destroyed. But the `is_destroyed` check on the current develop relies on the destruction of `singleton<Foo>` so the flag is set to false. As that destruction does not happen the flag stays true although `Foo` was destroyed. Accessing it (via `singleton<Foo>::get_const_instance()`) will then access invalid memory and crash (see #111)

For this the indirection via the `detail::singleton_wrapper` is required, which also allows usage of `Foo` with a protected ctor, (which is recommended: One should not be able to do e.g. `Foo f;`)

Note: The alternative is to force all users (`T`) to inherit from `singleton<T>` with the additional advantage that copy construction/assignment can be forbidden in `singleton<>` to avoid even more potential misuses.

The 2nd fix commit removes the assertions on `is_destroyed` before it is runtime checked. Due to the runtime check the assertion is not required and unless one can truly and reliably fix the shared-library-issue the runtime check will be required to avoid the crash. Having the assertion in place might make someone assume the runtime check can be removed. Obviously the assertion **will** fail for #111 after  `is_destroyed` is fixed making it 2 reasons to remove.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ install:
 script:
   - |-
     echo "using $TOOLSET : : $TRAVIS_COMPILER ;" > ~/user-config.jam
-  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared}
+  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared} cxxflags=-fPIC cflags=-fPIC
 
 notifications:
   email:
--- a/include/boost/serialization/singleton.hpp
+++ b/include/boost/serialization/singleton.hpp
@@ -58,7 +58,7 @@ namespace serialization {
 // details.
 //
 
-// singletons created by this code are guarenteed to be unique
+// Singletons created by this code are guaranteed to be unique
 // within the executable or shared library which creates them.
 // This is sufficient and in fact ideal for the serialization library.
 // The singleton is created when the module is loaded and destroyed
@@ -74,14 +74,21 @@ namespace serialization {
 // Second, it provides a mechanism to detect when a non-const function
 // is called after initialization.
 
-// make a singleton to lock/unlock all singletons for alteration.
+// Make a singleton to lock/unlock all singletons for alteration.
 // The intent is that all singletons created/used by this code
 // are to be initialized before main is called. A test program
-// can lock all the singletons when main is entereed.  This any
-// attempt to retieve a mutable instances while locked will
-// generate a assertion if compiled for debug.
+// can lock all the singletons when main is entered.  Thus any
+// attempt to retrieve a mutable instance while locked will
+// generate an assertion if compiled for debug.
 
-// note usage of BOOST_DLLEXPORT.  These functions are in danger of
+// The singleton template can be used in 2 ways:
+// 1 (Recommended): Publicly inherit your type T from singleton<T>,
+// make its ctor protected and access it via T::get_const_instance()
+// 2: Simply access singleton<T> without changing T. Note that this only
+// provides a global instance accesible by singleton<T>::get_const_instance()
+// To prevent using multiple instances of T make its ctor protected
+
+// Note on usage of BOOST_DLLEXPORT: These functions are in danger of
 // being eliminated by the optimizer when building an application in
 // release mode. Usage of the macro is meant to signal the compiler/linker
 // to avoid dropping these functions which seem to be unreferenced.
@@ -113,34 +120,69 @@ static inline singleton_module & get_singleton_module(){
     return m;
 }
 
+namespace detail {
+
+// This is the class actually instantiated and hence the real singleton.
+// So there will only be one instance of this class. This does not hold
+// for singleton<T> as a class derived from singleton<T> could be
+// instantiated multiple times.
+// It also provides a flag `is_destroyed` which returns true, when the
+// class was destructed. It is static and hence accesible even after
+// destruction. This can be used to check, if the singleton is still
+// accesible e.g. in destructors of other singletons.
+template<class T>
+class singleton_wrapper : public T
+{
+    static bool & get_is_destroyed(){
+        // Prefer a static function member to avoid LNK1179.
+        // Note: As this is for a singleton (1 instance only) it must be set
+        // never be reset (to false)!
+        static bool is_destroyed_flag = false;
+        return is_destroyed_flag;
+    }
+public:
+    singleton_wrapper(){
+        BOOST_ASSERT(! is_destroyed());
+    }
+    ~singleton_wrapper(){
+        get_is_destroyed() = true;
+    }
+    static bool is_destroyed(){
+        return get_is_destroyed();
+    }
+};
+
+} // detail
+
 template <class T>
 class singleton {
 private:
-    // note presumption that T has a default constructor
     static T & m_instance;
     // include this to provoke instantiation at pre-execution time
     static void use(T const &) {}
     static T & get_instance() {
-        static T t;
-
-        // refer to instance, causing it to be instantiated (and
-        // initialized at startup on working compilers)
         BOOST_ASSERT(! is_destroyed());
+
+        // use a wrapper so that types T with protected constructors can be used
+        // Using a static function member avoids LNK1179
+        static detail::singleton_wrapper< T > t;
 
         // note that the following is absolutely essential.
         // commenting out this statement will cause compilers to fail to
         // construct the instance at pre-execution time.  This would prevent
         // our usage/implementation of "locking" and introduce uncertainty into
-        // the sequence of object initializaition.
+        // the sequence of object initialization.
         use(m_instance);
 
         return static_cast<T &>(t);
     }
 
-    static bool & get_is_destroyed(){
-        static bool is_destroyed;
-        return is_destroyed;
-    }
+
+protected:
+    // Do not allow instantiation of a singleton<T>. But we want to allow
+    // `class T: public singleton<T>` so we can't delete this ctor
+    BOOST_DLLEXPORT singleton(){}
+
 public:
     BOOST_DLLEXPORT static T & get_mutable_instance(){
         BOOST_ASSERT(! get_singleton_module().is_locked());
@@ -150,16 +192,12 @@ public:
         return get_instance();
     }
     BOOST_DLLEXPORT static bool is_destroyed(){
-        return get_is_destroyed();
-    }
-    BOOST_DLLEXPORT singleton(){
-        get_is_destroyed() = false;
-    }
-    BOOST_DLLEXPORT ~singleton() {
-        get_is_destroyed() = true;
+        return detail::singleton_wrapper< T >::is_destroyed();
     }
 };
 
+// Assigning the instance reference to a static member forces initialization
+// at startup time as described in http://tinyurl.com/ljdp8
 template<class T>
 T & singleton< T >::m_instance = singleton< T >::get_instance();
 
--- a/src/extended_type_info.cpp
+++ b/src/extended_type_info.cpp
@@ -125,7 +125,6 @@ BOOST_SERIALIZATION_DECL void
 extended_type_info::key_unregister() const{
     if(NULL == get_key())
         return;
-    BOOST_ASSERT(! singleton<detail::ktmap>::is_destroyed());
     if(! singleton<detail::ktmap>::is_destroyed()){
         detail::ktmap & x = singleton<detail::ktmap>::get_mutable_instance();
         detail::ktmap::iterator start = x.lower_bound(this);
--- a/src/extended_type_info_typeid.cpp
+++ b/src/extended_type_info_typeid.cpp
@@ -95,7 +95,7 @@ BOOST_SERIALIZATION_DECL void
 extended_type_info_typeid_0::type_unregister()
 {
     if(NULL != m_ti){
-        BOOST_ASSERT(! singleton<tkmap>::is_destroyed());
+        // Check that the singleton is still alive (might be unloaded already)
         if(! singleton<tkmap>::is_destroyed()){
             tkmap & x = singleton<tkmap>::get_mutable_instance();
 
--- a/include/boost/archive/impl/archive_serializer_map.ipp
+++ b/include/boost/archive/impl/archive_serializer_map.ipp
@@ -47,10 +47,6 @@ archive_serializer_map<Archive>::insert(const basic_serializer * bs){
 template<class Archive>
 BOOST_ARCHIVE_OR_WARCHIVE_DECL void
 archive_serializer_map<Archive>::erase(const basic_serializer * bs){
-    BOOST_ASSERT(! boost::serialization::singleton<
-            extra_detail::map<Archive>
-        >::is_destroyed()
-    );
     if(boost::serialization::singleton<
         extra_detail::map<Archive>
     >::is_destroyed())
--- a/src/void_cast.cpp
+++ b/src/void_cast.cpp
@@ -276,7 +276,6 @@ void_caster::recursive_register(bool includes_virtual_base) const {
 
 BOOST_SERIALIZATION_DECL void
 void_caster::recursive_unregister() const {
-    BOOST_ASSERT(! void_caster_registry::is_destroyed());
     if(void_caster_registry::is_destroyed())
         return;
 
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -167,6 +167,8 @@ if ! $(BOOST_ARCHIVE_LIST) {
         [ test-bsl-run test_smart_cast ]
         [ test-bsl-run test_codecvt_null ]
         [ test-bsl-run test_singleton ]
+        [ test-bsl-run test_singleton_plain ]
+        [ test-bsl-run test_singleton_inherited ]
 
         # [ test-bsl-run test_z ]
 
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -52,6 +52,22 @@ lib dll_polymorphic_derived2
         <link>shared:<define>BOOST_SERIALIZATION_DYN_LINK=1 
     ;
 
+lib multi_shared_1
+    :
+        multi_shared1.cpp
+        ../build//boost_serialization/<link>static
+    :
+        <link>shared
+    ;
+
+lib multi_shared_2
+    :
+        multi_shared2.cpp
+        ../build//boost_serialization/<link>static
+    :
+        <link>shared
+    ;
+
 test-suite "serialization" :
      [ test-bsl-run_files test_array : A : :  [ requires cxx11_hdr_array ] ] # BOOST_NO_CXX11_HDR_ARRAY
      [ test-bsl-run_files test_boost_array : A ]
@@ -140,6 +156,8 @@ if ! $(BOOST_ARCHIVE_LIST) {
         [ test-bsl-run test_dll_simple   : : dll_a : ]
         # [ test-bsl-run test_dll_plugin : : polymorphic_derived2 : <link>static:<build>no <target-os>linux:<linkflags>-ldl ]
 
+        # Tests using shared libraries which are linked against static boost. Hence ignore for shared boost build
+        [ test-bsl-run test_multi_shared_lib : : multi_shared_1 multi_shared_2 : <link>shared:<build>no ]
         [ test-bsl-run test_private_ctor ]
         [ test-bsl-run test_reset_object_address : A ]
         [ test-bsl-run test_void_cast ]
--- a/test/test_singleton_inherited.cpp
+++ b/test/test_singleton_inherited.cpp
@@ -1,0 +1,82 @@
+/////////1/////////2/////////3/////////4/////////5/////////6/////////7/////////8
+// test_singleton_inherited.cpp:
+// Test the singleton class for a "inherited" singleton (used as Foo:public singleton<Foo>)
+// This can be uses as singleton<Foo>::get_const_instance() OR Foo::get_const_instance()
+//
+// - is_destroyed returns false when singleton is active or uninitialized
+// - is_destroyed returns true when singleton is destructed
+// - the singleton is eventually destructed (no memory leak)
+
+// (C) Copyright 2018 Alexander Grund
+// Use, modification and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include "test_tools.hpp"
+#include <boost/serialization/singleton.hpp>
+#include <boost/preprocessor/stringize.hpp>
+#include <stdexcept>
+
+// Can't use BOOST_TEST because:
+// a) destructors are called after program exit
+// b) This is intended to be used by shared libraries too which would then need their own report_errors call
+// We halso have to disable the Wterminate warning as we call this from dtors
+// C++ will terminate the program in such cases which is OK here
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wterminate"
+#define THROW_ON_FALSE(cond) if(!(cond)) throw std::runtime_error(__FILE__ "(" BOOST_PP_STRINGIZE(__LINE__) ") Assertion failed: " #cond)
+
+// Enum to designate the state of the singletonized instances
+enum ConstructionState{CS_UNINIT, CS_INIT, CS_DESTROYED};
+
+// We need another singleton to check for the destruction of the singletons at program exit
+// We don't need all the magic for shared library anti-optimization and can keep it very simple
+struct controller{
+    static controller& getInstance(){
+        static controller instance;
+        return instance;
+    }
+    ConstructionState state;
+private:
+    controller() {
+        state = CS_UNINIT;
+    }
+    ~controller();
+};
+
+// A simple class that sets its construction state in the controller singleton
+struct Foo: boost::serialization::singleton<Foo>{
+    Foo(): i(42) {
+        // access controller singleton. Therefore controller will be constructed before this
+        THROW_ON_FALSE(controller::getInstance().state == CS_UNINIT);
+        controller::getInstance().state = CS_INIT;
+    }
+    ~Foo() {
+        // Because controller is constructed before this, it will be destructed AFTER this. Hence controller is still valid
+        THROW_ON_FALSE(controller::getInstance().state == CS_INIT);
+        controller::getInstance().state = CS_DESTROYED;
+    }
+    // Volatile to prevent compiler optimization from removing this
+    volatile int i;
+};
+
+controller::~controller() {
+    // If this fails, the singletons were not freed and memory is leaked
+    THROW_ON_FALSE(state == CS_DESTROYED);
+    // If this fails, then the destroyed flag is not set and one may use a deleted instance if relying on this flag
+    THROW_ON_FALSE(boost::serialization::singleton<Foo>::is_destroyed());
+    THROW_ON_FALSE(Foo::is_destroyed());
+}
+
+int
+test_main( int /* argc */, char* /* argv */[] )
+{
+    // Check if the singleton is alive and use it
+    BOOST_CHECK(!boost::serialization::singleton<Foo>::is_destroyed());
+    BOOST_CHECK(!Foo::is_destroyed());
+
+    BOOST_CHECK(boost::serialization::singleton<Foo>::get_const_instance().i == 42);
+    BOOST_CHECK(Foo::get_const_instance().i == 42);
+    return EXIT_SUCCESS;
+}
+
--- a/test/test_singleton_plain.cpp
+++ b/test/test_singleton_plain.cpp
@@ -1,0 +1,78 @@
+/////////1/////////2/////////3/////////4/////////5/////////6/////////7/////////8
+// test_singleton_plain.cpp:
+// Test the singleton class for a "plain" singleton (used as singleton<Foo>)
+//
+// - is_destroyed returns false when singleton is active or uninitialized
+// - is_destroyed returns true when singleton is destructed
+// - the singleton is eventually destructed (no memory leak)
+
+// (C) Copyright 2018 Alexander Grund
+// Use, modification and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include "test_tools.hpp"
+#include <boost/serialization/singleton.hpp>
+#include <boost/preprocessor/stringize.hpp>
+#include <stdexcept>
+
+// Can't use BOOST_TEST because:
+// a) destructors are called after program exit
+// b) This is intended to be used by shared libraries too which would then need their own report_errors call
+// We halso have to disable the Wterminate warning as we call this from dtors
+// C++ will terminate the program in such cases which is OK here
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wterminate"
+#define THROW_ON_FALSE(cond) if(!(cond)) throw std::runtime_error(__FILE__ "(" BOOST_PP_STRINGIZE(__LINE__) ") Assertion failed: " #cond)
+
+// Enum to designate the state of the singletonized instances
+enum ConstructionState{CS_UNINIT, CS_INIT, CS_DESTROYED};
+
+// We need another singleton to check for the destruction of the singletons at program exit
+// We don't need all the magic for shared library anti-optimization and can keep it very simple
+struct controller{
+    static controller& getInstance(){
+        static controller instance;
+        return instance;
+    }
+    ConstructionState state;
+private:
+    controller() {
+        state = CS_UNINIT;
+    }
+    ~controller();
+};
+
+// A simple class that sets its construction state in the controller singleton
+struct Foo{
+    Foo(): i(42) {
+        // access controller singleton. Therefore controller will be constructed before this
+        THROW_ON_FALSE(controller::getInstance().state == CS_UNINIT);
+        controller::getInstance().state = CS_INIT;
+    }
+    ~Foo() {
+        // Because controller is constructed before this, it will be destructed AFTER this. Hence controller is still valid
+        THROW_ON_FALSE(controller::getInstance().state == CS_INIT);
+        controller::getInstance().state = CS_DESTROYED;
+    }
+    // Volatile to prevent compiler optimization from removing this
+    volatile int i;
+};
+
+controller::~controller() {
+    // If this fails, the singletons were not freed and memory is leaked
+    THROW_ON_FALSE(state == CS_DESTROYED);
+    // If this fails, then the destroyed flag is not set and one may use a deleted instance if relying on this flag
+    THROW_ON_FALSE(boost::serialization::singleton<Foo>::is_destroyed());
+}
+
+int
+test_main( int /* argc */, char* /* argv */[] )
+{
+    // Check if the singleton is alive and use it
+    BOOST_CHECK(!boost::serialization::singleton<Foo>::is_destroyed());
+
+    BOOST_CHECK(boost::serialization::singleton<Foo>::get_const_instance().i == 42);
+    return EXIT_SUCCESS;
+}
+


Flamefire:
Rebased on develop and shows that both failures from #129 and #128 are fixed by this.

As this PR includes the other 2 it can simply be merged instead of them or after the other 2.

src/extended_type_info.cpp
```diff
@@ -125,7 +125,6 @@ BOOST_SERIALIZATION_DECL void
 extended_type_info::key_unregister() const{
     if(NULL == get_key())
         return;
-    BOOST_ASSERT(! singleton<detail::ktmap>::is_destroyed());
```
jeking3: The original author was suggesting that if the singleton is destroyed here, then something was quite wrong.  An invariant was broken somehow.  I agree with the original author on this one.  Why would the map have been destroyed AND something is calling unregister?  It sounds like there's a static uninitialization race somewhere.  Isn't this assertion still valid?  If not, there's still a race.

.travis.yml
```diff
@@ -96,7 +96,7 @@ install:
 script:
   - |-
     echo "using $TOOLSET : : $TRAVIS_COMPILER ;" > ~/user-config.jam
-  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared}
+  - ./b2 -j 3 libs/serialization/test toolset=$TOOLSET link=${LINK:-shared} cxxflags=-fPIC cflags=-fPIC
```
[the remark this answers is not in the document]
Flamefire: Because I need this when building static boost but link it against dynamic libs to reproduce this. For static boost bjam does not do this

src/extended_type_info_typeid.cpp
```diff
@@ -95,7 +95,7 @@ BOOST_SERIALIZATION_DECL void
 extended_type_info_typeid_0::type_unregister()
 {
     if(NULL != m_ti){
-        BOOST_ASSERT(! singleton<tkmap>::is_destroyed());
+        // Check that the singleton is still alive (might be unloaded already)
         if(! singleton<tkmap>::is_destroyed()){
```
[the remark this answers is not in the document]
Flamefire: See #128:

- Assume 2 libraries using `extended_type_info_typeid_0` (singletons) with different types. Lets call these singletons `type<Foo>` and `type<Bar>`
- On initialization time both types register themselves in `singleton<tkmap>`, which hence gets constructed before.
  - Note that each library has its own instance of `singleton<tkmap>` due to static linkage of boost
- On destruction time library A gets unloaded: `type<Foo>` gets destructed, unregisters itself in `singleton<tkmap>`, afterwards `singleton<tkmap>` gets destructed.
- Due to those destruction issue in shared libs **both** `singleton<tkmap>` get destructed (you can check this easily with PRINTF in #128)
- Now when `type<Bar>` gets destroyed the `singleton<tkmap>` it wants to access is already destroyed
  - Previously this caused a BOOM -> See #128
  - Now the assertion would fire and the runtime check catches this.
  - No unregistering is required as the register does not exist altogether.
  - All good, just skip the unregistering code

src/extended_type_info.cpp
Flamefire: Yes there is a broken invariant: With shared libs using static boost as described in the other occurence you commented on. Cause is simply implementation defined behaviour for shared libs (see difference Linux vs Windows and OSX)

So no the assertion is not valid anymore hence its removal. This condition is instead caught by the runtime check 1 line below.

Its is not a race as the destruction happens on unload in single-threaded way.

jeking3:
@Flamefire thanks for the answers.  I didn't pick that up from the previous conversations.

Flamefire:
Sure. I suggest also reading the description of https://github.com/boostorg/serialization/pull/105 and #104 for more details. But I think the important parts are now summarized here. Thanks for asking.

jeking3:
UBSAN identified some issues in the area - this one happens many times during the tests:
```
testing.capture-output ../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.run
====== BEGIN OUTPUT ======
../../boost/serialization/singleton.hpp:175:12: runtime error: reference binding to null pointer of type 'const struct map'
    #0 0x7f5cc1f16b95 in boost::serialization::singleton<boost::archive::detail::extra_detail::map<boost::archive::binary_iarchive> >::get_instance() ../../boost/serialization/singleton.hpp:175
    #1 0x7f5cc1f0ebd4 in __static_initialization_and_destruction_0 ../../boost/serialization/singleton.hpp:202
    #2 0x7f5cc1f0ebf1 in _GLOBAL__sub_I_binary_iarchive.cpp ../../libs/serialization/src/binary_iarchive.cpp:40
    #3 0x7f5cc20e42d9  (/lib64/ld-linux-x86-64.so.2+0x102d9)
    #4 0x7f5cc20e43c2  (/lib64/ld-linux-x86-64.so.2+0x103c2)
    #5 0x7f5cc20d5299  (/lib64/ld-linux-x86-64.so.2+0x1299)


EXIT STATUS: 1
====== END OUTPUT ======

    LD_LIBRARY_PATH="/home/travis/build/jeking3/boost-root/bin.v2/libs/filesystem/build/gcc-8/debug/cxxstd-03-iso/visibility-hidden:/home/travis/build/jeking3/boost-root/bin.v2/libs/serialization/build/gcc-8/debug/cxxstd-03-iso/visibility-hidden:/home/travis/build/jeking3/boost-root/bin.v2/libs/system/build/gcc-8/debug/cxxstd-03-iso/visibility-hidden:/usr/bin:/usr/lib:/usr/lib32:/usr/lib64:$LD_LIBRARY_PATH"
export LD_LIBRARY_PATH

    status=0
    if test $status -ne 0 ; then
        echo Skipping test execution due to testing.execute=off
        exit 0
    fi
     "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive"   > "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.output" 2>&1 < /dev/null
    status=$?
    echo >> "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.output"
    echo EXIT STATUS: $status >> "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.output"
    if test $status -eq 0 ; then
        cp "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.output" "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.run"
    fi
    verbose=0
    if test $status -ne 0 ; then
        verbose=1
    fi
    if test $verbose -eq 1 ; then
        echo ====== BEGIN OUTPUT ======
        cat "../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.output"
        echo ====== END OUTPUT ======
    fi
    exit $status

...failed testing.capture-output ../../bin.v2/libs/serialization/test/test_diamond_xml_archive.test/gcc-8/debug/cxxstd-03-iso/visibility-hidden/test_diamond_xml_archive.run...
```

jeking3:
The valgrind build looks like it has identified some other things, but I don't see anything related to singletons or static construction or destruction.  

https://api.travis-ci.org/v3/job/451983024/log.txt

The windows builds don't look good either, but it may not be related to these changes.

jeking3:
On the windows side this is appearing in all of the builds so far, but it may not be related to the singleton changes.
```
msvc.link.dll bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll
   Creating library bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.lib and object bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.exp
polymorphic_derived2.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl polymorphic_base::~polymorphic_base(void)" (__imp_??1polymorphic_base@@UEAA@XZ) referenced in function "public: virtual __cdecl polymorphic_derived2::~polymorphic_derived2(void)" (??1polymorphic_derived2@@UEAA@XZ)
polymorphic_derived2.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl polymorphic_base::polymorphic_base(void)" (__imp_??0polymorphic_base@@QEAA@XZ) referenced in function "public: static void __cdecl boost::serialization::access::construct<class polymorphic_derived2>(class polymorphic_derived2 *)" (??$construct@Vpolymorphic_derived2@@@access@serialization@boost@@SAXPEAVpolymorphic_derived2@@@Z)
polymorphic_derived2.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl polymorphic_base::polymorphic_base(class polymorphic_base const &)" (__imp_??0polymorphic_base@@QEAA@AEBV0@@Z) referenced in function "public: __cdecl polymorphic_derived2::polymorphic_derived2(class polymorphic_derived2 const &)" (??0polymorphic_derived2@@QEAA@AEBV0@@Z)
polymorphic_derived2.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: class polymorphic_base & __cdecl polymorphic_base::operator=(class polymorphic_base const &)" (__imp_??4polymorphic_base@@QEAAAEAV0@AEBV0@@Z) referenced in function "public: class polymorphic_derived2 & __cdecl polymorphic_derived2::operator=(class polymorphic_derived2 const &)" (??4polymorphic_derived2@@QEAAAEAV0@AEBV0@@Z)
bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll : fatal error LNK1120: 4 unresolved externals
        call "bin.v2\standalone\msvc\msvc-14.1\address-model-64\architecture-x86\msvc-setup.bat"  >nul
 link /NOLOGO /INCREMENTAL:NO /DLL /MACHINE:X64 /MANIFEST /subsystem:console /out:"bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll" /IMPLIB:"bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.lib"    @"bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll.rsp"
        if %ERRORLEVEL% NEQ 0 EXIT %ERRORLEVEL%
    
...failed msvc.link.dll bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.lib...
...removing bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.lib
```
https://ci.appveyor.com/project/jeking3/serialization/build/job/344fh48thj6hs66f?fullLog=true#L3555

robertramey:
Note regarding some UBSAN errors/warnings.  The serialization library in some cases depends on undefined behavior in order to function. In particular with the serialization of objects through pointers.  In particular, particular in a diamond inheritance situation.  Some of these may be "fixable" but it's possible that some may not and we'll just have to depend on testing to verify that the code actually works.  These will need to be considered individually.

Flamefire:
This particular UB is probably caused by the required instantiation force thingy @pdimov mentioned https://github.com/boostorg/serialization/pull/105#discussion_r225619041

I argued that this should be done in a seaparate PR to not mix things up.

include/boost/serialization/singleton.hpp
```diff
@@ -150,16 +192,12 @@ class singleton {
...
-    BOOST_DLLEXPORT ~singleton() {
-        get_is_destroyed() = true;
+        return detail::singleton_wrapper< T >::is_destroyed();
     }
 };
 
+// Assigning the instance reference to a static member forces initialization
+// at startup time as described in http://tinyurl.com/ljdp8
 template<class T>
 T & singleton< T >::m_instance = singleton< T >::get_instance();
```
jeking3: Why do this at all?  On the first call to get_instance() for a type, it would instantiate the class when needed.

Flamefire: Because it is needed to be instantiated at startup time (requirement of Boost.Serialization)

This is checked by the `locked` feature https://github.com/boostorg/serialization/pull/131/files#diff-3f1e861be15fd4926f5e5b00388f2309R188

include/boost/serialization/singleton.hpp
```diff
@@ -113,34 +120,69 @@ static inline singleton_module & get_singleton_module(){
...
     static T & get_instance() {
-        static T t;
-
-        // refer to instance, causing it to be instantiated (and
-        // initialized at startup on working compilers)
         BOOST_ASSERT(! is_destroyed());
 
+        // use a wrapper so that types T with protected constructors can be used
+        // Using a static function member avoids LNK1179
+        static detail::singleton_wrapper< T > t;
```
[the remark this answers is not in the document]
Flamefire: Those lines are required so the initialization-at-startup works. https://github.com/boostorg/serialization/pull/131/files#diff-3f1e861be15fd4926f5e5b00388f2309R160 and https://github.com/boostorg/serialization/pull/131/files#diff-3f1e861be15fd4926f5e5b00388f2309R202 should suffice but they need the `use` call or it will be optimized away.

include/boost/serialization/singleton.hpp
jeking3: I see, the requirements are to avoid locking so this needs to be done at initialization time.  This is a pretty funky way to do it. :)

Flamefire: That's why I added the comment and a reference so future readers know why this is done. I deem this an important part of this PR

.travis.yml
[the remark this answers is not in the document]
robertramey: FWIW - I never add the flags -fPIC.  I "trust" bam to "do the right thing" for the platform and configuration.  This could well be responsible for getting different test results.  In my view if these are necessary, they should be added to the the appropriate jam script for all shared libraries.  Then there is the issue of what the "link=shared" switch actually does.  It's my belief that it would link to the shared version of the C++ runtime library.  But then there are also C++ system libraries like libc++ - a different thing.  Then there are one's own shared libraries as we are using in our tests here.  So what does link=shared/static apply to?  As far as I'm concerned it's never been explained.  Oh and don't forget the run-link feature.  How should it be set in what circumstances? It goes on an on and on and on.  B2 build system injects myriad unexplained side-effects into one's code.  Worse these side-effects may vary silently from platform to platform.  It also doesn't help at all that the boost test matrix doesn't make it easy to see how these things are set for each test.  All this makes it very, very, very difficult to know that your build is what you really want.  And makes it much, much more difficult to collaborate given that there's no easy way to guarantee that differing testers are actually testing the same thing.

src/extended_type_info_typeid.cpp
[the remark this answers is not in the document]
robertramey: I expect that most singletons used in the serialization library to be constructed pre-main time and so in these cases I would expect is_destroyed to always return false.  BUT questions arise regarding this for dynamically loaded shared libraries.  Each shared library probably has it's own set of singletons.  Since shared libraries are not addressed by the standard and behavior is not documented anywhere that I know of, I don't know this for sure.  But it seems like the only reasonable way to do it.  In such a case I can see how a single might be created / destroyed before/after main is called.  Still since usage of such a singleton would be confined to the interior of the library, things should be OK.  Of course if the same type is used (maybe only declared in a header) things could get more complex - maybe two singletons of the same type might exist???.   Because of all this - and because I'm paranoid in general, I created is_destroyed basically to support runtime asserts and to gate things which might happen and might be legal.  It's basically defensive programming.  Down side is that it's really hard to guarantee that is_destroyed has the value that I expect of it.  In a simple case where types aren't dynamically loaded - I feel OK about this.  In other cases - it's much harder to say.  Given that all this depends so much on other/unknowable things, I don't think anyone can say much more than this.   We're stuck with dependence on testing.

src/extended_type_info.cpp
robertramey: Hmmm - as the original author, I think my intention is pretty clear.  So far I haven't managed to trip this assertion.  If does occur then I'd like to know about it.  Now I'm aware that you've got some tests which which do trip this assertion.  When I've looked at your tests before I've not been convinced that they prove what you want to prove.  Having said that, I'm willing to look at more tests.  But I would like to see them more carefully constructed.  I'm aware that this requires a lot more effort but I don't see any other choice.

include/boost/serialization/singleton.hpp
robertramey: Note on locking.  Strictly speaking this "locking" feature is not necessary.  In the serialization library I wanted all my singletons to be constant - ie not mutable.  I wanted to trap attempts to modify them after main is called.  But of course that means I want to force creation on program/shared library load.  So I'm fine with the comment.  I would recommend that the comment not point to a link - even if the text is longer.  The serialization library will likely persist in usage long after the link is gone.

src/extended_type_info.cpp
pdimov: > So far I haven't managed to trip this assertion.

That's because `is_destroyed` doesn't work.

.travis.yml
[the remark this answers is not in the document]
pdimov: > FWIW - I never add the flags -fPIC. I "trust" bam to "do the right thing" for the platform and configuration.

Unfortunately, `bjam` doesn't do the right thing when building shared libraries that incorporate static libraries. It adds `-fPIC` when building the shared library, but not when building the static library.

> Then there is the issue of what the "link=shared" switch actually does. It's my belief that it would link to the shared version of the C++ runtime library.

No, that's `runtime-link=shared`. `link=shared` applies to the libraries built by `b2`, makes them shared.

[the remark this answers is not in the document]
robertramey: off topic - so what is runtime-link set to by default?  My understanding is that at least on windows if one module is set to use static version of the C++ runtime and another is set to use Dll  version, ceos ensues. as far as I know, there's no obstacle to one doing this.  b2 is opaque to me.

and BTW - it seems that -fPIC shouldn't be necessary when building a static library as the addresses are fixed up at link time.

src/extended_type_info.cpp
robertramey: Hmmm - how does it not work.
`
    static bool & get_is_destroyed(){
        static bool is_destroyed;
        return is_destroyed;
    }
public:
    BOOST_DLLEXPORT static T & get_mutable_instance(){
        BOOST_ASSERT(! get_singleton_module().is_locked());
        return get_instance();
    }
    BOOST_DLLEXPORT static const T & get_const_instance(){
        return get_instance();
    }
    BOOST_DLLEXPORT static bool is_destroyed(){
        return get_is_destroyed();
    }
    BOOST_DLLEXPORT singleton(){
        get_is_destroyed() = false;
    }
    BOOST_DLLEXPORT ~singleton() {
        get_is_destroyed() = true;
    }
`

where is the problem here?  I understand that the singleton destructor might not ever be called as it would happen post main time.  But that's not in anyone's control.  I don't know if one would say that that's an error.  In any case we don't depend on that and have no control over it anyway. The intention is to detect that case where some singleton is being uses after it's destructor has been called.  So to say it "doesn't work" would mean that ~singleton has been called but the is_destroyed still returns true.  I don't see any evidence of that happening.  Of course if assertions are being invoked - it would mean that singleton destructors are not being called in reverse sequence of the constructors.  The library is predicated on the assumption that does not happen.  Soooo, if this assertion is being invoked I would say that it's not an indication that the is_destroyed "doesn't work", it's more likely that the assumption of sequence of destruction/construction is not respected.  Of course this could be the scenario which might occur when one is invoking singletons from multiple dynamically loaded shared libraries.  Which I believe is the situtation we are trying to deal with in this case.  If this is indeed occurring - then the true question is - under what conditions does this occur and how can it be prevented.  

In other words, I believe that if the assertions are being invoked - that's it's a true violation of some assumption that I've made.  Not necessarily that the conditional is wrong.

pdimov: The constructor and destructor of `singleton<detail::ktmap>` are never called, because a variable of such type is never created.

.travis.yml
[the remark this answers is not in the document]
pdimov: The default for `runtime-link` is `shared`. `b2` will not let you mix the two. `-fPIC` is necessary when building a static library if you later incorporate this library as part of a shared library.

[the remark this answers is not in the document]
Flamefire: @ `-fPIC`: The crash happens when static boost is linked into shared libs. Hence:

- Build a new library: https://github.com/boostorg/serialization/pull/131/files#diff-e97299ad55a4013dc19889043aa464a5R55
- Make it a shared library: https://github.com/boostorg/serialization/pull/131/files#diff-e97299ad55a4013dc19889043aa464a5R60
- Link against static boost: https://github.com/boostorg/serialization/pull/131/files#diff-e97299ad55a4013dc19889043aa464a5R58
- Due to this (static lib is **consumed** by dynamic lib which is not "backpropagated by bjam) add -fPIC

src/extended_type_info.cpp
Flamefire: > I understand that the singleton destructor might not ever be called as it would happen post main time.

That's where you are mislead: The whole situation is about what happens "post main time": When all the singletons gets destructed and types get unregistered a mixup occurs and things go BOOM.

I described the test I did in #129 here: https://github.com/boostorg/serialization/pull/131#discussion_r232009933. I argue that this is a very carefully constructed test as it completely relies on the C++ standard. It shows under what circumstances the problem (`is_destroyed` returns false for a destroyed singleton) occurs: When the class is not inherited from singleton<> as it is the case for `singleton<detail::ktmap>` as @pdimov  also noticed. (There are 2 tests: 1 where it is inherited and 1 where it is not, both are valid as per your docu and code) This is why the subclass `singleton_wrapper` is required: To listen to the destruction of `detail::ktmap`, not (only) `singleton<>`

> If this is indeed occurring - then the true question is - under what conditions does this occur and how can it be prevented.

The first part can be verified easily: Cherry-pick https://github.com/boostorg/serialization/pull/131/commits/7a34ef8b76e542560ac3895b5a3d68499f7ed422 and run the test #128 on linux in debug mode: The assertion **will** fire. I do not think this can be prevented because it is "implementation defined land" (shared libs), but it can be gracefully handled (by the runtime checks)

Side note: What do you think is wrong with #128? I made this out of #111 (which is the minimum working example) with completely documented code. No tricks or anything. Or did I miss anything?

include/boost/serialization/singleton.hpp
Flamefire: I actually think it is required as a test that construction happens before main. If it would be constructed on first use, you won't be able to deserialize a derived class from a pointer for which you didn't "use" the singleton yet (e.g. by serializing it or refering to it otherwise)

I can remove the link if you wish or replace it by the original: https://groups.google.com/forum/#!topic/microsoft.public.vc.language/kDVNLnIsfZk (in the hope that even in the unlikely case google will vanish services as archive.org can be used to retrieve it), but the essentials are summarized in the 1 sentence already. Your call.

src/extended_type_info.cpp
robertramey: >The constructor and destructor of singleton<detail::ktmap> are never called, because a variable of such type is never created.
Why do you think that? 
a) Certainly the get_instance which creates the singleton<detail::ktmap> is getting called.  Otherwise things would not work at all.  So the constructor must be getting called.  And there is a static instance created at pre-compile time so the constructor must be getting called there.
b) I can't guarantee that the destructor is getting at post-main time as I don't have much access to it.
c) Note that there is a test called "test_singleton" which verifies that singleton constructors are called and that destructors are called in reverse order.  See https://github.com/boostorg/serialization/blob/develop/test/test_singleton.cpp

Now the question arises as to what happens when we've got multiple singletons of the same type. Such as in different shared libraries.  It might be possible that there would be some sort of mixup in that results in the wrong destructor being called.  But I don't see an easy way to track this down.

robertramey: >That's where you are mislead: The whole situation is about what happens "post main time": When all the singletons gets destructed and types get unregistered a mixup occurs and things go BOOM.

That's conceivable. I'll take another look at your test.  I'll look at "test_plain" as it seems to be simpler.

>The first part can be verified easily: Cherry-pick 7a34ef8 and run the test #128 on linux in debug mode: The assertion will fire. I do not think this can be prevented because it is "implementation defined land" (shared libs), but it can be gracefully handled (by the runtime checks)

Right - I don't have a linux system setup up for this.  But I'll still review the test.  But I want to review the best test so if you think there's a better one than #129/test_plain let me know.

If your assessment of the problem above is correct, it should manifest itself replacing singleton<detail::ktmap> with some simpler type - Correct.  The reason I ask this is that I'd like to verify that it's an issue with singleton and something else in the serialization library.  Perhaps you'd like to make that change and report the results.

If you assessment above is correct, then we need a workaround.  The workaround is not the complex in theory - just suppress the assertion at post main time. I'm sure a way to do that could be found.  But of course everything is trickier than it looks.  In particular, what is post-main time when a dll is unloaded?  Is it really post main time when things go south - or is it at dll unload time? etc. etc....

pdimov: > Certainly the `get_instance` which creates the `singleton<detail::ktmap>` is getting called. Otherwise things would not work at all. So the constructor must be getting called.

`singleton<detail::ktmap>::get_instance` is a static member. It creates an instance of `detail::ktmap` and returns it. This however has nothing to do with the constructor of `singleton<detail::ktmap>`.

pdimov: In the case where `X` derives from `singleton<X>`, constructing a `X` necessarily constructs the base `singleton<X>`. `detail::ktmap`, however, is a typedef for `std::multimap<>`, which does not derive from `singleton<>`.

This is all explained in Alexander's test which checks both cases (he calls them "plain" singleton and "inherited" singleton.)

robertramey: > This however has nothing to do with the constructor of singleton<detail::ktmap>

Hmmm - does this not invoke the constructor of singleton<detail::ktmap> at pre-main time?

template<class T>
T & singleton< T >::m_instance = singleton< T >::get_instance();

pdimov: No; `T` here is `detail::ktmap`, so it defines a variable of type `detail::ktmap&`, initialized with the result of calling `singleton<detail::ktmap>::get_instance()`. Nothing here constructs a `singleton<detail::ktmap>`.

robertramey: OK - I see this now and it's not what I intended.  I'll look into this

robertramey: OK - I'm buying this.  I'll put in my current version and run some tests on my machine then check it in.  I did change the m_instance from a reference to a pointer to avoid an error with UBSAN set.  Hopefully that won't break anything.

robertramey:
msvc.link.dll bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.dll
   Creating library bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.lib and object bin.v2\libs\serialization\test\msvc-14.1\release\address-model-64\threading-multi\dll_polymorphic_derived2-vc141-mt-x64-1_69.exp
polymorphic_derived2.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl polymorphic_base::~polymorphic_base(void)" (__imp_??1polymorphic_base@@UEAA@XZ) referenced in function "public: virtual __cdecl polymorphic_derived2::~polymorphic_derived2(void)" (??1polymorphic_derived2@@UEAA@XZ)
...

This is not related to the "singleton" problem.  In working to find the error in test_dll_exported on some platforms (which I thought was related to the "singleton" problem), I discovered an instance where mixing exported types serialized via base class pointers to polymorphic archives (LOL - now that is a mouthful) wouldn't link.  This turns out to be trickier than I thought as I had to navigate the "visibility/import/export" labyrinth.  I've got it fixed on platforms other than windows which I cannot test here.  So this is still pending. 

As an aside - the test_dll_exported test failure on some very few platforms is also related to "visibility"  This could be verified by building with visibility=global.  This whole episode wasn't made any easier by the fact that global build with b2 changed the default visibility rules while I was doing this and for a while maintained different settings between develop and master.  Very, very, very confusing and hard to deal with.

jeking3:
I have fixes pending for the polymorphic.  They are building right now in https://github.com/jeking3/serialization/pull/4

They are mixed in with better CI job definitions, and I have enabled the compilation of all the examples as well.

robertramey:
I'm running the singleton fixes now my machine.  Probably take a number of hours.  I still have pending questions regarding the -fPIC switch with b2 which I would like to see considered in a more general way. I expect UBSAN to show errors regardless of what I do.  This should eliminate one source of these errors but I'm pretty sure they can never be totally eliminated.

pdimov:
> I still have pending questions regarding the -fPIC switch with b2 which I would like to see considered in a more general way.

`-fPIC` should be a propagated feature, but that needs to be implemented on Boost.Build's side. I mentioned that on the list - https://lists.boost.org/Archives/boost/2018/11/244239.php - but the more appropriate place would probably be an issue in Build.

src/extended_type_info.cpp
Flamefire: @pdimov already said the vital parts but I still want to answer the following to avoid future problems:

> But I want to review the best test so if you think there's a better one than #129/test_plain let me know.

Actually this is answered in the description of #129: There are 2 possible use cases for singleton. `extended_type_info` uses the inherited-paradigm and `tkmap` the plain-paradigm. For both `is_destroyed` has to work. That's why **both**  tests are required. So please do not only merge one of them or it will be broken again.

Flamefire:
@robertramey What do you think of adding `-fPIC` to the bjam file manually: https://github.com/jeking3/serialization/pull/3/commits/3bf664304b14ba1ca55d15d79ae4d4fdae9e2b13#diff-42dd6ec1330a7c47aaccf2ab2b8f1e02

robertramey:
>What do you think of adding -fPIC to the bjam file manually

I'm pretty confused about the PIC thing.  I understand what it's for.  But in order to be changing only one thing at a time - I didn't include that change.  I think I remember reading that without PIC the addresses are fixed up when the shared library is loaded in to the main programs address space.  I don't know if that's actually true.  I'm sure Peter will have something to say about this.  In any case, it seems that all the tests that might be related to singleton are passing.  If it's not actually necessary, I'd prefer to not include it as it injects another confusing variable into the mix.  The only failure is test_dll_exported which is failing only under windows.  I'm quite sure this is related to visibility.

Right now, I'm trying to get the last bug out so I can merge to master so it can come out in the soon to be released version.  I'm already overdue and that's a problem.

I've take a cursory look at your tests and they look OK.  Still I want to take a little more time and before I add them to the official test suite.  So they wouldn't go in until after we get this merge done.

Flamefire:
> But in order to be changing only one thing at a time - I didn't include that change

Completely agreed. That's why I made separate PRs and even commits for that.

>  I think I remember reading that without PIC the addresses are fixed up when the shared library is loaded in to the main programs address space.

"PIC" = "Position Independent Code". A shared library may be loaded anywhere in the address space and hence the generated assembly must be aware of this and include relative references, fixups and the like. From a more practical point of view you can simply say: All object files linked into a shared library must be build with `-fPIC`, which includes static libs. You get a linker error telling you this.

>  If it's not actually necessary, I'd prefer to not include it as it injects another confusing variable into the mix. 

I do think the test of #128 is valuable enough to avoid regressions. The bjam modification is generic enough to be included

> I've take a cursory look at your tests and they look OK. Still I want to take a little more time and before I add them to the official test suite. So they wouldn't go in until after we get this merge done.

With the time pressure and the critical issue fixed by this I do agree the merge to master is most important. Just don't forget those tests (especially the "easier" #129) afterwards to avoid this in the future. I chose this kinda TDD approach so this issue **stays** fixed.

jeking3:
On the polymorphic changes, I have something that passes more tests, but it may not be what was originally intended.  What I noticed is that the polymorphic_derived2 class was being declared as an export, but it was importing the polymorphic_base.  Is the intention to compile the test code and the polymorphic_base into the test EXE, and to compile the polymorphic_derived2 into the DLL?  If so, my changes aren't correct.  I modified it so that the base has the same linkage as the derived.  I think more work is needed here.

Fork PR: https://github.com/jeking3/serialization/pull/4
AppVeyor passes all MSVC tests now (but still has warnings like:
```
compile-c-c++ bin.v2\libs\serialization\test\test_exported_polymorphic_text_archive.test\msvc-14.1\release\address-model-64\threading-multi\test_exported.obj
test_exported.cpp
polymorphic_derived2 imported
c:\projects\boost-root\libs\serialization\test\polymorphic_derived2.hpp(50): warning C4275: non dll-interface class 'polymorphic_base' used as base for dll-interface class 'polymorphic_derived2'
c:\projects\boost-root\libs\serialization\test\polymorphic_base.hpp(41): note: see declaration of 'polymorphic_base'
c:\projects\boost-root\libs\serialization\test\polymorphic_derived2.hpp(48): note: see declaration of 'polymorphic_derived2'
compile-c-c++ bin.v2\libs\serialization\test\test_exported_polymorphic_text_archive.test\msvc-14.1\release\address-model-64\threading-multi\polymorphic_base.obj
```
):
https://ci.appveyor.com/project/jeking3/serialization/builds/20164701

Travis CI has UBSAN and Valgrind failures but everything else passes:
https://travis-ci.org/jeking3/serialization/builds/452693625

Flamefire:
Nice, but as this is an unrelated issue can you open another PR and discuss it there?

jeking3:
Let's discuss it on the fork PR.  Good suggestion.

.travis.yml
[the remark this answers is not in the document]
res2k: I like to add `-fPIC` via a build feature, it's more "surgical" and automatic (you need to pass correct flags on bjam command line).
See here: https://github.com/res2k/serialization/commit/b279e155bb9ba18499c38daa0dfe3a190a23390d

robertramey:

PhilippMuenzel:
How do I find out which PR has been merged to the official boost release version? In other words, is this included in the boost 1.69 release or will it be?

jeking3:
It appears all commits in develop are in master, so changes related to this (these exact fixes were not necessarily applied, as the PR was not merged) were likely committed.

PhilippMuenzel:
thank you 👍

PR #139, opened 2018-11-13
Fix Windows shared linking tests

res2k:
* Only tag symbols as dll-exported when building a DLL
* Only tag symbols as dll-imported when using a DLL
* Don't define DLL-imported method
* Add missing dependency

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -46,10 +46,16 @@ lib dll_polymorphic_base
 
 lib dll_polymorphic_derived2
     :
+        dll_polymorphic_base
+
         polymorphic_derived2.cpp
         ../build//boost_serialization
     : 
         <link>shared:<define>BOOST_SERIALIZATION_DYN_LINK=1 
+        <link>shared:<define>POLYMORPHIC_DERIVED2_DYN_LINK=1
+    :
+    :
+        <link>shared:<define>POLYMORPHIC_DERIVED2_DYN_LINK=1
     ;
 
 test-suite "serialization" :
--- a/test/polymorphic_derived2.hpp
+++ b/test/polymorphic_derived2.hpp
@@ -24,7 +24,7 @@
 #include <boost/serialization/type_info_implementation.hpp>
 #include <boost/serialization/extended_type_info_typeid.hpp>
 
-#if defined(BOOST_ALL_DYN_LINK) || defined(BOOST_SERIALIZATION_DYN_LINK)
+#if (defined(BOOST_ALL_DYN_LINK) || defined(BOOST_SERIALIZATION_DYN_LINK)) && defined(POLYMORPHIC_DERIVED2_DYN_LINK)
     #if defined(POLYMORPHIC_DERIVED2_IMPORT)
         #define POLYMORPHIC_DERIVED2_DLL_DECL BOOST_SYMBOL_IMPORT
         #pragma message ("polymorphic_derived2 imported")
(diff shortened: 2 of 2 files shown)


robertramey:
I wasn't crazy about your method here of adding yet another macro.  I resolved this by slightly restructuring code to better separate declaration from definition.

res2k:
I added Yet Another Macro to make it work on MinGW as well.

The MinGW linker doesn't like it if you use a symbol with a `dllimport` attribute in one object, but with a `dllexport` attribute in another; the latter doesn't define the `__imp_*` symbols that are expected.
(The MSVC linker sucks this up, but emits a warning LNK4217.)
Thus you need to have an `dllexport` attribute for on the symbol definition that goes into the DLL, and _no_ `dll*` attribute for the object file that will be used "statically". (Or have the symbol also appear as `dllexport` in places it's used as well; that's how you use a `dllexport` symbol from the DLL itself.)

FWIW, the typical approach for DLLs on Windows is having one `FOO_DECL` (resolving to `dllexport` or `dllimport` as appropriate) macro per library. That's the robust way to get the correct `dll` attributes. Trying to cover multiple libraries correctly while only using a single macro is hard, if not impossible.

robertramey:
hmm - I'm thinking you're on to something - but I need a better explanation to understand this.

I presume you're referring to the module named dll_polymorphic_derived2.

>Thus you need to have an dllexport attribute for on the symbol definition that goes into the DLL, and no dll* attribute for the object file that will be used "statically". (Or have the symbol also appear as dllexport in places it's used as well; that's how you use a dllexport symbol from the DLL itself.)

I'm thinking that this is a quirk of the gcc compiler.  Generally it does not require "visibility" for imported symbols declared in a header.  It only requires visibility when those symbols exported from another module.   As you note - windows does it differently. So wouldn't the "correct" way to handle this to be that for this platform - mingw - , POLYMORPHIC_BASE_DLL_DECL be defined as nothing for imported symbols and BOOST_SYMBOL_EXPORT for exported symbols?

robertramey:
Hmm - I'm surprised that removing the dllimport on the calling program doesn't break the compile on msvc platforms.  To see the whole point was to only be able to invoke functions that one explicitly imported. I'm very aware of the MS dll/com/... import export thing going back more that 20 years.  It's a ill thought out hack and has always been so.  The gcc - visibility isn't much better.  And of course the C++ committee spends all it's time on stuff other than things like this which are and always have been a problem.

But enough of the rant.  You've demonstrated the issue.  But I think the fix in in the wrong place.  Fixing in the test is "hiding the bug".  It makes the test pass, but it doesn't help for the next time.  Think bigger! Take a look at this excerpt from polymorphic_derived2.hpp:

`#if defined(BOOST_ALL_DYN_LINK) || defined(BOOST_SERIALIZATION_DYN_LINK)
    #if defined(POLYMORPHIC_DERIVED2_IMPORT)
        #define POLYMORPHIC_DERIVED2_DLL_DECL BOOST_SYMBOL_IMPORT
        #pragma message ("polymorphic_derived2 imported")
    #elif defined(POLYMORPHIC_DERIVED2_EXPORT)
        #define POLYMORPHIC_DERIVED2_DLL_DECL BOOST_SYMBOL_EXPORT
        #pragma message ("polymorphic_derived2 exported")
    #endif
#endif
`
To me it seems that the real solution is that the value of BOOST_SYMBOL_IMPORT should be defined to nothing for the MINGW platform.  Following up on this will almost surely lead to config.hpp.  This is John Maddox's territory.  I asked him about this on either the user or developer's list and he was at a loss.  So I'd suggest that you finish nailing this down and propose a definitive fix at the source.  I'm sure he'd be appreciative.  And you'd have the satisfaction of not fixing one test in the serialization library, but fixing for now and forever any boost application which uses boost, mingw and DLL.  That's what I mean by thinking bigger!

res2k:
Hmm, one issue would be that `dllimport` is required for variables. Just setting `BOOST_SYMBOL_IMPORT` to empty would instantly break any users of variables across modules. Sure, you can work around that by having a function returning a reference, but that may penalize other platforms.

Also, an empty `BOOST_SYMBOL_IMPORT` would inhibit the supposed efficiency gain on Windows, however small that is. I could imagine that fear of a performance decline would have some users prefer a retention of `dllimport`.

So, an empty `BOOST_SYMBOL_IMPORT` would probably not be an universal improvement.

robertramey:
> So, an empty BOOST_SYMBOL_IMPORT would probably not be an universal improvement.

If you look at how config.hpp works, you'll see that these macros are defined by the combination of compiler and the OS.  So one would expect to "fine tune" the definition of BOOST_SYMBOL_IMPORT so that it would be defined as dll_import on plain windows and leave it undefined on MINGW.  This of course would be the optimal solution.  As the issue that windoze/mingw makes it even more complex by having different rules for imported variables than for imported functions ... what can I say.  Nothing surprise me any more.  In any case, of hand I don't know of a way to handle this.

PR #143, opened 2019-01-04
more robust setting of optional version

dkoes:
I think it would be better to use the macro provided in version.hpp since otherwise the version<boost::optional<T>> struct is missing fields (e.g. tag).

### DIFF
--- a/include/boost/serialization/optional.hpp
+++ b/include/boost/serialization/optional.hpp
@@ -96,12 +96,9 @@ void serialize(
     boost::serialization::split_free(ar, t, version);
 }
 
-template<class T>
-struct version<boost::optional<T> > {
-    BOOST_STATIC_CONSTANT(int, value = 1);
-};
-
 } // serialization
 } // boost
+
+BOOST_CLASS_VERSION(boost::optional< T >, 1)
 
 #endif // BOOST_SERIALIZATION_OPTIONAL_HPP_


robertramey:
what is the motivation for this?

dkoes:
Really obtuse compiler errors when trying to serialize without versioning information.

```
#include <boost/optional.hpp>
#include <boost/serialization/utility.hpp>
#include <boost/serialization/optional.hpp>
#include <boost/archive/binary_oarchive.hpp>
#include <boost/archive/text_oarchive.hpp>

#include <iostream>
using namespace boost;
namespace boost {
namespace serialization {

//default all our classes to not have version info
template<class T>
struct implementation_level_impl<const T> {
    template<class U>
    struct traits_class_level {
        typedef BOOST_DEDUCED_TYPENAME U::level type;
      };

      typedef mpl::integral_c_tag tag;

      typedef
      BOOST_DEDUCED_TYPENAME mpl::eval_if<
      is_base_and_derived<boost::serialization::basic_traits, T>,
      traits_class_level< T >,
      //else
      BOOST_DEDUCED_TYPENAME mpl::eval_if<
      is_fundamental< T >,
      mpl::int_<primitive_type>,
      //else
      BOOST_DEDUCED_TYPENAME mpl::eval_if<
      mpl::or_<is_class< T >, is_array< T> >,
      mpl::int_<object_serializable>,
      //else
      BOOST_DEDUCED_TYPENAME mpl::eval_if<
      is_enum< T >,
      mpl::int_<primitive_type>,
      //else
      mpl::int_<not_serializable>
      >
      >
      >
      >::type type;
      BOOST_STATIC_CONSTANT(int, value = type::value);
    };


}
}

//BOOST_CLASS_VERSION(boost::optional< short unsigned int  >, 0)

int main() {
   boost::archive::detail::check_object_versioning< boost::optional<short unsigned int> >();

}
```

robertramey:
first of all, this change generates failed tests on all configurations and all platforms.

second, something like BOOST_CLASS_VERSION(boost::optional< T >, 1) doesn't work as the < and > characters really mess up the intent.

The original code using struct ... works around the problem above.  

So this change is ill conceived on multiple levels.  I'm going to close it without merging it in.

I hope I'm discourage any future efforts at improving/fixing the library.  I depend on users such as yourself to help me out.  But not everyone gets it right all the time.

Thanks again.

Robert Ramey

PR #148, opened 2019-03-11
Add support for C++17 std::variant

### LINKED ISSUE #114: add seralization for std::optional and std::variant
We need a serialization function for std::optional and std::variant.
Maybe they can be placed in `boost/serialization/optional.hpp`, the file for boost::optional and `boost/serialization/variant.hpp` for boost::variant, or new header files?

### LINKED ISSUE #121: Serialization of boost::optional should use load_construct_data
It currently isn't possible to serialize a boost::optional&lt;T&gt; where T doesn't have a default constructor. This used to work, but was changed in 632df7aab5e1fb3458ea53be2704c83462b2caa9. However, this is an important use case. While the workaround is easy, there is usually a good reason not to allow a default constructor, for example there isn't a reasonable default value. In other cases, it may not even be possible to add a default constructor, for example when adding serialization to 3rd party classes. 

I can't think for a good reason for this limitation, as the serialization framework defines the load_construct_data hook specifically for this use case. Since boost::optional can be considered a container with either 0 or 1 items, I would therefore expect it to also be serialized in a similar way to the STL container, and call load_construct_data (using use stack_construct&lt;T&gt; I assume).

sdebionne:
Partially fixes #114 and drops support for MSVC6.


robertramey:
the normal way to do this is to include a line in the Jamfile which looks like the following:

     [ test-bsl-run_files test_array : A : :  [ requires cxx11_hdr_array ] ] # BOOST_NO_CXX11_HDR_ARRAY

This will enable the test for all compilers which support std::array.  This depends in boost config defining BOOST_NO_CXX11_HDR_ARRAY.  So in your case you'll need to use BOOST_NO_CXX11_HDR_VARIANT.  But I see that boost config doesn't include this yet.  So you'll have to bug John Maddock to get this included.  They you'll be good to go.  BTW - since you're on this, you might want to look at boost variant2 - in development by peter dimov

sdebionne:
AFAIU, `variant2` is a drop in replacement of `std::variant` for C+11 so it should work out of the box (beside the namespace), are you suggesting to add support for this as well?

I try to add a property `cxxstd=17` in the test requirements of the variant test, but that gives `error: 'cxxstd=17' is not a valid property specification`. What would be the right way to switch to C++17 for a single test?

sdebionne:
@robertramey The only test (test_polymorphic_helper_polymorphic_binary_archive) that do not pass actually fails on the develop branch to. Any idea what is wrong?

robertramey:
> @robertramey The only test (test_polymorphic_helper_polymorphic_binary_archive) that do not pass actually fails on the develop branch to. Any idea what is wrong?

 Notice that it only happens with gcc on windows (mingw?), with polymorphic_binary_archive, on static linking.  The executable seems to run but then ... the expected *.test file isn't found.  I'm guessing that this is indicates some obscure issue with the jam file in the test directory.   I haven't been able to find it.  The only real way to would be to test on a local machine with that configuration - which I don't have. 

So I haven't worried about this failure.  If you wanted to spend some time on this obscure circumstance, that would be great.  But you can ignore it as well.

p12tic:
@robertramey @sdebionne : Is there any way I could help to land this PR? Is it that test failure that prevents merging or something else? `std::variant` support would be really useful for me. Thanks!

include/boost/serialization/std_variant.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+#include <boost/serialization/split_free.hpp>
+#include <boost/serialization/serialization.hpp>
+#include <boost/serialization/nvp.hpp>
+
+namespace boost {
+namespace serialization {
+
+template<class Archive>
+struct variant_save_visitor
```
p12tic: This will clash with a class with the same name on boost::variant's variant.hpp

include/boost/serialization/std_variant.hpp
```diff
@@ -0,0 +1,199 @@
...
+    ){
+        if(which == 0){
+            // note: A non-intrusive implementation (such as this one)
+            // necessary has to copy the value.  This wouldn't be necessary
+            // with an implementation that de-serialized to the address of the
+            // aligned storage included in the variant.
+            using type = mp::front<Seq>;
+            type value;
+            ar >> BOOST_SERIALIZATION_NVP(value);
+            v = value;
```
p12tic: We could use `std::move` to allow non-copy-constructible types. The serialization support for boost::variant has the same issue though.

include/boost/serialization/std_variant.hpp
```diff
@@ -0,0 +1,199 @@
...
+    struct typelist {};
+
+    template <typename Seq>
+    using front = typename detail::front_impl<Seq>::type;
+
+    template <typename Seq>
+    using pop_front = typename detail::pop_front_impl<Seq>::type;
+}  // namespace mp
+
+template<std::size_t N, class Seq>
```
p12tic: N looks like unused...

sdebionne:
@p12tic Thanks for the review, I think all your points are valid, I'll fix that ASAP.

Regarding the merge, I won't have time to investigate the (unrelated) pipeline failure now, but I think @robertramey was OK to merge as is.

Boost.Variant2 is landing in 1.71. Since it has the same interface than the `std`, I wonder if we could factorize both serializations here.

As a side note, I have an `std::optional` serialization to, but will open a separate PR not to delay this one.

include/boost/serialization/std_variant.hpp
sdebionne: `N` is used for recursion, see line 134: `variant_impl<N - 1, types>::load(ar, which - 1, v, version);`

p12tic: Ah, I missed that the types parameter is the same.

include/boost/serialization/std_variant.hpp
```diff
@@ -0,0 +1,199 @@
...
+    std::size_t which;
+    ar >> BOOST_SERIALIZATION_NVP(which);
+    if(which >=  sizeof...(Types))
+        // this might happen if a type was removed from the list of variant types
+        boost::serialization::throw_exception(
+            boost::archive::archive_exception(
+                boost::archive::archive_exception::unsupported_version
+            )
+        );
+    variant_impl<sizeof...(Types), mp::typelist<Types...>>::load(ar, which, v, version);
```
pdimov: I'd use mp11 here, specifically `mp_with_index` on `which`.

robertramey: Hmmm - this would add mp11 as a dependency of the serialization library.  I'd be reluctant to do this to accommodate one data type.  I understand that it wouldn't preclude the rest of the library from continue to be C++03 compliant. But the net result would be a pretty confusing situation where the level of compatibility would depend upon which types one is serializing.

I've always advocated that serialization of boost types be specified in the particular boost library itself.  This eliminates the above problem.  But then library authors (justifiable) object to introduction of a dependency of their library on serialization.  The real solution to this conundrum is to complete the factorization of the serialization library into the two halves serialization and archive.  Turns out that most of this is done as I was aware of this from the beginning.  But I gave it up near the end as it was getting difficult to maintain while actually getting all the functionality that boost required to be inserted into the library.  Undertaking this task now would be possible, not require wholesale changes, but would be intricate and run the risk of introducing subtle problems and be somewhat tedious to undertake.

mateuszzz88:
I think you should add serialization of `std::monostate` - it is equivalent of `boost::blank`
```
namespace boost {
namespace serialization {

template<class Archive>
void serialize(Archive &ar, std::monostate &, const unsigned int /*version*/)
{}

}
}
```
I see no reason to put in separate file, variant.hpp seems ok (monostate is in the same header `<variant>`).

sdebionne:
@pdimov Do you see any way I could support Boost.Variant2 without duplicating the code too much?

pdimov:
If you remove the std:: from std::visit and std::get, the same code should also work for variant2. Instead of variant_size you can use mp_size which should work on both (but you'll need to remove qualifiers and references if present.) Also, what I said above about mp11::mp_with_index.

robertramey:
> Instead of variant_size you can use mp_size which should work on both (but you'll need to remove qualifiers and references if present.) Also, what I said above about mp11::mp_with_index.

Hmmm - this would bump the Compiler requirement from C++03 to C++11.   But then it would only do so for those using std::variant or boost::variant2.  So I'd have to think about this.

robertramey:
Whoops.  I misspoke and I apologize.  I got the notice of a commit and didn't realize it was for your PR and not the boost develop repo.  Chalk it up to some GitHub confusion.

kilasuelika (approved):
I have test `std_variant`, it works perfectly.

code-affinity:
When I try to compile this header in a context in which <boost/serialization/variant.hpp> has already been included, there is a name conflict between the variant_impl struct that is declared in this header, and the variant_impl struct that is declared in <boost/serialization/variant.hpp>

I resolved the problem by changing the three occurrences of variant_impl to stdvariant_impl.

Febbe:
@robertramey whats missing here? May you overtake this PR when it does not fulfill your expectations?

Issue #152, opened 2019-04-06
Minor CMake path issue


nemothenoone:
Does the path at https://github.com/boostorg/serialization/blob/develop/CMake/CMakeLists.txt#L103 is missing an extra ```..```?

I mean, if I use CMake project to build Boost.Serialization inside Boost superproject, I get ```BOOST_ROOT not found```, but extra ```..``` fixes the problem.

robertramey:
Hmmmm ...  I've hacked my CMake implementation to support my personal debug/development of Boost serialization with my Xcode IDE.  Previously I used the mdvc IDE which was fine to do the development and debug, but hell to maintain whenever anything (boost version, msvc version, etc).  CMake was sort of pain to figure out - it's better now - but it was a one time deal.  Now whenever something changes I can just regenerate a new Xcode (and presumably MSVC or Eclipse or .. ) project and I'm back in business.  It only takes a minute.  Better I can maintain projects for different IDE simultaneously using the same source code base.  Presumable this opens the path to building the boost build type library with a single CMake command - though I've never tried this - so its unlikely to work with out some more tweaking.  There are several things that you might want to support: Building from the Cmake command line, generate correctly named boost libraries and dlls, adding support for permitting the the serialization library to be consumed as a CMake like component, supporting CDash and perhaps CMake style deployment.  As to your particular case, I'm not sure what you're actually doing so I don't know what to suggest.  All I can say is that it works for me for what I use it for.

If you're using my CMake files for something different than I am you might need to tweak something.  If you find the magic enhancement - and it doesn't break it for the purposes I use it for, post it as a PR and I'll seriously consider adding it in.

PR #153, opened 2019-04-08
Minor CMake path changes

nemothenoone:
Pull request for the issue #152.

Minor CMake enhancements and fixes are about strict standard-compliant option type (```INTEGER``` -> ```STRING```) and path fixes for CMake project being built inside of boost superproject source tree.

Try it yourself. Clone ```https://github.com/boostorg/boost``` and ```cmake -G Xcode ${BOOST_ROOT}/libs/serialization/CMake``` from some directory. Without this minor fix, it leads me to the error like ```CMake Error at CMakeLists.txt:106 (message): BOOST_ROOT not found```

### DIFF
--- a/CMake/CMakeLists.txt
+++ b/CMake/CMakeLists.txt
@@ -95,12 +95,12 @@ endif()
 #
 
 set(Boost_DEBUG true)
-set(Boost_ADDRESS_MODEL 64 CACHE INTEGER "32/64 bits")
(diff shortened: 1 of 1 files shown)


robertramey:
making the change to line 103, adding another level of ..,  results in my getting BOOST_ROOT not found error.  We're out of sync somehow.

PR #158, opened 2019-05-12
fix missing header and inverted ifdef

HDembinski:
@glenfe asked me to support BOOST_NO_EXCEPTIONS in Boost.Histogram and thus I found this issue. The serialization version of throw_exception calls ::boost::throw_exception but didn't include the header <boost/throw_exception.hpp>, so it only worked by chance if that header was already included.

### DIFF
--- a/include/boost/serialization/throw_exception.hpp
+++ b/include/boost/serialization/throw_exception.hpp
@@ -17,8 +17,9 @@
 
 #include <boost/config.hpp>
 
-#ifndef BOOST_NO_EXCEPTIONS
+#ifdef BOOST_NO_EXCEPTIONS
 #include <exception>
+#include <boost/throw_exception.hpp>
 #endif
 
 namespace boost {


HDembinski:
@robertramey Could you please have a look at the failing test? I looked into the log and I can't find out what's wrong with. It would be really nice if you could accept this PR. I am using Boost.Serialization in Boost.Histogram and I cannot test the serialization parts with exceptions turned off because of this bug.

HDembinski:
On second look, I am not sure whether my patch is correct.

robertramey:
FYI - My personal experience is that running tests on one's own system is much better than depending on CI tests.  For one things, one can debug the stuff on one's own system.  CI works best when you know that things work on your own system but you need confirmation that it works on other systems that you can test on.

HDembinski:
I do both. The CI tests make sure that I am not breaking things on Windows (I develop on Linux and OSX) and under special conditions that I do not test regularly, such as -fno-exceptions.

PR #160, opened 2019-06-04
Add preliminary support for generic serialization

sdebionne:
See #113 for the rationale.

Let me know if there is a better way to properly test for std=c++11 and the availability of <type_traits>.

Is there any other place in the library where functionalities are conditionally supported given a c++ standard level?


sdebionne:
@HDembinski @robertramey Any chance that you guys find time to review this PR? If accepted, it would  be great to have in 1.71...

include/boost/serialization/serialization.hpp
```diff
@@ -60,6 +64,44 @@ namespace serialization {
...
+
+struct has_serialize
+{
+
+private:
+    template<class T>
+    static constexpr auto check(T*)
+        -> typename
+        std::is_same<
+            decltype(std::declval<T>().serialize(std::declval<Archive&>(), std::declval<unsigned int>())),
```
HDembinski: Where is Archive& coming from here?

include/boost/serialization/serialization.hpp
```diff
@@ -60,6 +64,44 @@ namespace serialization {
 
 BOOST_STRONG_TYPEDEF(unsigned int, version_type)
 
+#if __cplusplus==201103L
```
HDembinski: Use detection from boost.config instead of raw checks like.

include/boost/serialization/serialization.hpp
```diff
@@ -13,6 +13,10 @@
 #include <boost/config.hpp>
 #include <boost/serialization/strong_typedef.hpp>
 
+#if !defined(BOOST_NO_CXX11_HDR_TYPE_TRAITS)
```
HDembinski: You don't need <type_traits>, versions of enable_if and is_same are in boost.core.

include/boost/serialization/serialization.hpp
```diff
@@ -60,6 +64,44 @@ namespace serialization {
...
+        -> typename
+        std::is_same<
+            decltype(std::declval<T>().serialize(std::declval<Archive&>(), std::declval<unsigned int>())),
+            void
+        >::type;
+
+    template<class>
+    static constexpr std::false_type check(...);
+
+    typedef decltype(check<C>(0)) type;
```
HDembinski: Perhaps Boost.Concept should be used instead of writing this manually.

HDembinski:
I am not the maintainer of this library, but here is a review.

HDembinski:
The whole check can be implemented in pre-C++11 code, I think. https://stackoverflow.com/questions/257288/is-it-possible-to-write-a-template-to-check-for-a-functions-existence

robertramey:
> Let me know if there is a better way to properly test for std=c++11 and the availability of <type_traits>.

Take serious look at Boost config.  It has been the cornerstone of boost serialization portability.  People take boost serialization portability for granted, but it's in fact an incredible feat which is totally due to the efforts of John Maddock's Boost config.  I never test for any specific C++ version - only the existence of a capability as specified by boost config.  It has to be this way because compilers which purport to support C++11 are not all in sync which combination of C++11 features they support.  The fact is that there is no defacto C++ standard - only a theoretical one.

> Is there any other place in the library where functionalities are conditionally supported given a c++ standard level?

easy. look through the bjam test files and the tests themselves.

HDembinski:
@robertramey I share your admiration of boost.config :)

PR #162, opened 2019-06-07
just forward throw_exception

HDembinski:
This very simple implementation fixes the problem I have in Boost.Histogram when I try to compile my unit tests which use Boost.Serialization with `<exception-handling>off`. It also simplifies this code. `boost::serialization::throw_exception` is just forwarded to `boost::throw_exception`, which already calls `throw` if BOOST_NO_EXCEPTION is not defined and a user-defined `throw_exception` function otherwise.

### DIFF
--- a/include/boost/serialization/throw_exception.hpp
+++ b/include/boost/serialization/throw_exception.hpp
@@ -15,28 +15,14 @@
 // accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
-#include <boost/config.hpp>
-
-#ifndef BOOST_NO_EXCEPTIONS
-#include <exception>
-#endif
+#include <boost/throw_exception.hpp>
 
 namespace boost {
 namespace serialization {
 
-#ifdef BOOST_NO_EXCEPTIONS
-
-BOOST_NORETURN inline void throw_exception(std::exception const & e) {
+template<class E> BOOST_NORETURN inline void throw_exception(E const & e){
     ::boost::throw_exception(e);
 }
-
-#else
-
-template<class E> BOOST_NORETURN inline void throw_exception(E const & e){
-    throw e;
-}
-
-#endif
 
 } // namespace serialization
 } // namespace boost


HDembinski:
@robertramey Now I am pretty sure the patch is correct, but the gcc-based appveyor builds still fail. But I don't see where. When my builds fail, I search for "error " but there is no error. I cannot find what is actually failing.

robertramey:
Yep - it's a bitch.  Here's what I do:
a) I got the page with the failed test.  In this case it's https://ci.appveyor.com/project/RobertRamey/serialization-5b0xm/builds/25116591/job/i98kmjbhtf0qpg71
b) I search the log for the text "...fail".  
c) this should show the failing test.

I think it might display only the first 2000 lines of the log.  In this case you have to download the log onto your machine so you can search it.

BTW - this is the same test/configuration which has been failing forever with dynamic linking on windows.  But I've never been able to correct it as I don't have a windows system either! With your change, it seems that the issues shows up with static linking as well.

> ****

glenfe:
@robertramey the issues you're seeing on Appveyor (Windows) might have to do with long paths for the generated test binaries that exceed some path limit.
```
[00:11:52] The system cannot find the path specified.
[00:11:52] The system cannot find the path specified.
[00:11:52] The system cannot find the path specified.
[00:11:52] The system cannot find the path specified.
```

glenfe:
I'll help out with the Windows GCC Appveyor test issues after this NVP change is done.

glenfe:
https://github.com/glenfe/boost.serialization/blob/develop/appveyor.yml succeeds with all compilers (GCC from Cygwin, GCC from Mingw, MSVC versions 2010 to 2017). It only test the default static linking, but you can add variants for shared linking.

robertramey:
FWIW - I've fixed the long test name which was triggering the bogus failure.  So at least that's address.  I haven't looked at throw exception in a long time.  I'm pretty sure I included the code to avoid calling a "upgrade" to boost exception which drew in all kinds of unexpected behavior and dependencies.  If I get some time I might revisit this decision.

PR #163, opened 2019-07-05
Add serialization for std::optional

### LINKED ISSUE #114: add seralization for std::optional and std::variant

sdebionne:
Fixes #114 together with #148.

### DIFF
--- a/include/boost/serialization/std_optional.hpp
+++ b/include/boost/serialization/std_optional.hpp
@@ -1,0 +1,99 @@
+#ifndef BOOST_SERIALIZATION_STD_OPTIONAL_HPP_
+#define BOOST_SERIALIZATION_STD_OPTIONAL_HPP_
+
+#if defined(_MSC_VER)
+# pragma once
+#endif
+
+/////////1/////////2/////////3/////////4/////////5/////////6/////////7/////////8
+// optional.hpp - non-intrusive serialization of optional types
+//
+// copyright (c) 2019 Samuel Debionne, ESRF
+//
+// Use, modification and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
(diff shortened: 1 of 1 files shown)


robertramey:
I don't see any tests for this.  That every "built-in" type which includes serialization code must be tested.  That is must have a test program written for it.  It's not hard - look at the rest of them.  But it is a requirement.  Looking forward to this.

sdebionne:
Speaking of the tests, the CI is broken since Dec 16, 2020.  The std::variant has tests but was never merged. So why should I bother?

Ulysses1337:
What I hope to do one day is replace all my Boost optionals with std. Can we name the bool field "initialized" for compatibility?

robertramey:
I merged this and pulled into my local repo.  I was considering writing a test for it.  Upon closer examination I found there is an implementation for boost::optional serialization which is almost identical to your implementation of serialization for std::optional. So it seems to me that we should include serialization for both boost::optional and std::option in the same file.  And the test for optional should be extended to include std::optional as well.  This will diminish the code which needs to be maintained in the future.

As an aside, I think you should have credited the original author of the serialization for boost::optional in your header.

sdebionne:
Of course it is almost identical as `boost` and `std`  `optionals` have the same interfaces.

> So it seems to me that we should include serialization for both boost::optional and std::option in the same file

Why force people that use `std::optional` to depend on `boost::optional` via this shared header?

> the test for optional should be extended to include std::optional as well

Agreed but given that `std::variant` did not get merged, I did not want to spend time on adding the test here.

> As an aside, I think you should have credited the original author of the serialization for boost::optional in your header.

OK. I would create a PR for that but since it has been reverted in #299, I won't.

robertramey:
I've incorporated your working boost/serialization/optional and boost/serialization/variant.  I think you'll find this satisfactory.  Feel free to review it and ask me about it.

PR #166, opened 2019-07-24
Let serialization/optional.hpp use boost::serialization::access, fix #165

### LINKED ISSUE #165: Serialization of optional does not use private default constructor.
The `friend class boost::serialization::access;` is ignored during serialization of optional. The private default constructor is not used. Following code shows the problem directly:

```
// g++ -o serialize_optional serialize_optional.cpp -lboost_serialization

#define BOOST_OPTIONAL_CONSTRUCTOR_ACCESS

#include <fstream>
#include <iostream>
#include <boost/serialization/nvp.hpp>
#include <boost/serialization/serialization.hpp>
#include <boost/archive/xml_iarchive.hpp>
#include <boost/archive/xml_oarchive.hpp>
#include <boost/serialization/access.hpp>
#include <boost/archive/text_oarchive.hpp>
#include <boost/archive/text_iarchive.hpp>
#include <boost/serialization/optional.hpp>
#include <boost/optional.hpp>

class Foo {
	public:
		Foo(int k) : i(k) {}
		friend std::ostream& operator<<(std::ostream& os, Foo const& foo) {
			return os << "Foo { " << foo.i << " }";
		}
	private:
		Foo() : i(-1) {};
		friend class boost::serialization::access;

		template<class Archive>
		void serialize(Archive & ar, const unsigned int){
			#define N(a) BOOST_SERIALIZATION_NVP(a)
			ar & N(i);
			#undef N
		};
		int i;
};

int main() {
	using Data = boost::optional<Foo>;
	std::cout << "Boost version: " << BOOST_VERSION << "\n";
	{
		auto f = Data(2345);
		std::ofstream os("test.xml");
		boost::archive::xml_oarchive oa(os);
		oa << boost::serialization::make_nvp("f", f);
	}
	{
		Data g;
		std::ifstream is("test.xml");
		boost::archive::xml_iarchive ia(is);
		ia >> boost::serialization::make_nvp("f", g);
		
		std::cout << g.get() << "\n";
	}
}
```

cosurgi:
This merge request fixes #165 , the original code is retained. To enable new behavior use `#define BOOST_OPTIONAL_CONSTRUCTOR_ACCESS`.

Unfortunately I am not sure how to check for existence of private default constructor, so I left this check empty.

### DIFF
--- a/include/boost/serialization/access.hpp
+++ b/include/boost/serialization/access.hpp
@@ -17,6 +17,11 @@
 //  See http://www.boost.org for updates, documentation, and revision history.
 
 #include <boost/config.hpp>
+#if ! defined(BOOST_NO_CXX11_HDR_TYPE_TRAITS)
+    #include <type_traits>
+#else
+    #include <boost/type_traits/has_trivial_constructor.hpp>
+#endif
 
 namespace boost {
 
@@ -137,6 +142,26 @@ public:
     static T * cast_pointer(U * u){
         return static_cast<T *>(u);
     }
+
+    template <class X, class = decltype(X())>
+    static std::true_type test_any_constructor(X*);
+
+    template <class X>
+    static std::false_type test_any_constructor(...);
+
+    template<class T>
+    struct is_default_constructible_not_publicly {
+        static constexpr bool value
+            = decltype(access::test_any_constructor<T>(0))::value
+              and (not
+            #if ! defined(BOOST_NO_CXX11_HDR_TYPE_TRAITS)
+                    std::is_default_constructible<T>::value
+            #else
+                    boost::has_trivial_constructor<T>::value
+            #endif
+              );
+    };
+
 };
 
 } // namespace serialization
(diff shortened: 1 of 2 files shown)


cosurgi:
@robertramey I managed to detect if there is a not public default constructor. So now we have old (fast) behavior if there is a public constructor and the workaround behavior if the constructor is private. So no loss in performance.

Now the only problem is that it works only in C++17. We need to find a way to make it work on older compilers.

cosurgi:
I have a different idea to get back to the "old" serialization::load pre-1.65 code and adjust it to rely on access::construct (via stack_construct, like other serializable types) before "ar >> ....", thus fixing the problem of serializing into un-initialized object and also fixing my problem. My classes rely on "friend class boost::serialization::access" to ensure that only serialization code is able to use the default constructor and I really want to stick to this design decision, since it prevents bugs where objects might be otherwise constructed with no parameters.

I will create another pull request for this solution.

PR #173, opened 2019-09-04
Use boost::nvp from Boost.Core

glenfe:
All your tests pass on my Travis queue: https://travis-ci.org/glenfe/boost.serialization/builds/580758159

### DIFF
--- a/include/boost/serialization/nvp.hpp
+++ b/include/boost/serialization/nvp.hpp
@@ -16,51 +16,7 @@
 
 //  See http://www.boost.org for updates, documentation, and revision history.
 
-//////////////////// boost/core/nvp.hpp Starts here
-#include <utility> // pair
-
-#include <boost/config.hpp>
-#include <boost/detail/workaround.hpp>
-
-#include <boost/core/addressof.hpp>
-
-namespace boost {
-
-template<class T>
-struct nvp : 
-    public std::pair<const char *, T *>
-    // ,public wrapper_traits<const nvp< T > >
-{
-//private:
-    nvp(const nvp & rhs) :
-        std::pair<const char *, T *>(rhs.first, rhs.second)
-    {}
-public:
-    explicit nvp(const char * name_, T & t) :
-        // note: added _ to suppress useless gcc warning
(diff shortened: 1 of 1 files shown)


glenfe:
The Appveyor failures with GCC static are not related to this change.

robertramey:
I'm OK with this - but I'd like to spend a little time thinking about it before I merge it.  As I've said - I think that shuffling around components to address the dependency "problem" is not an effective strategy or even a good idea.  Never the less, I've been outvoted on this so I'm inclined to go along at least until I can create a comprehensive solution to the problem - which I know the outlines of but haven't had time to work on.  It's low priority since it won't be popular either.

glenfe:
I know, and agree. In the interim  - until your solution is completed and realized - with the way things are: This would enable a more than a few things to move forward and make progress (all without affecting existing users).

glenfe:
Robert, thanks for getting back to us on this on Slack quickly. We replied in the channel, but to not lose your attention while we have it: We think `boost::nvp` (https://www.boost.org/doc/libs/develop/libs/core/doc/html/core/nvp.html) having that one `serialize()` member is OK, given that:
* `nvp` is defined to be "Serializable"
* `nvp`'s `serialize()` specification is simple and uncoupled from anything in the Serialization library.

We could make it an undocumented member in the specification of `boost::nvp`. We could even remove the member entirely, and define `serialize()` for `boost::nvp` in Boost.Serialization. (If you prefer the last approach, I'm happy to update this pull request accordingly).

This doesn't make things worse, in any way, though. Users continue to use `boost::serialization::make_nvp` as before, or new users use `boost::make_nvp`, and all that has changed is in which library the type lives. It would be not much different to, say, `std::pair` being supported by Serialization but living outside the library.

It shouldn't inhibit any plans you have for improving Serialization, including future designs that make it easier to write "serializable" types without a dependency on Serialization. In the meanwhile, this will do is both allow us to make progress on two fronts: The modularization efforts, and other libraries (like MultiArray, and CircularBuffer) supporting Boost.Serialization serializability as early as the 1.72 release.

PR #175, opened 2019-10-01
Use make_nvp from Core

glenfe:
After this is merged this, we will update Core so that `make_nvp` is also defined in namespace `boost::serialization`.

### DIFF
--- a/include/boost/serialization/nvp.hpp
+++ b/include/boost/serialization/nvp.hpp
@@ -40,11 +40,7 @@ namespace boost {
 namespace serialization {
 
 using boost::nvp;
-
-template<class T>
-const nvp< T > make_nvp(const char * name, T & t){
-    return nvp< T >(name, t);
-}
+using boost::make_nvp;
 
 template<class Archive, class T>
 void save(


robertramey:
Hmmmm - shouldn't 
using boost::make_nvp;
using boost::nvp;

Both be eliminated as both nvp and make_nvp are now going to be in the serialization namespace?

glenfe:
They _can_ be eliminated, yes. However, the reason we chose to put them in `boost::serialization` in Boost.Core is to avoid breaking users who forward declare them. But the documentation of Boost.Core will specify that they are available only from `boost::` (the fact that they are within `boost::serialization` is an implementation detail that we do not want to guarantee to new users).

In Serialization, you have the choice to rely on that implementation detail, because we would never change it, but it would be nice if your view of them were also as Boost.Core facilities that you're bringing into Serialization.

glenfe:
@robertramey Note that until you merge this pull request I cannot move `make_nvp` into the `serialization` namespace, because all the tests would fail for all libraries using Serializaiton. Because then there would be two 'make_nvp' functions defined inside `boost::serialization`.

glenfe:
I merged https://github.com/boostorg/core/commit/e94af0d41a31c994967e45f18e2503e51c44c537

PR #198, opened 2020-04-20
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74

eldiener:


robertramey:
where is BOOST_EMBTC defined and how is it different from BOOST_BORLAND?

PR #209, opened 2020-06-24
another header fix

HDembinski:
The previous fix was not complete, another missing header was required.


robertramey:
should be included by now -

PR #230, opened 2021-02-28
Guard MSVC pragmas from clang-cl

pdimov:
This patch avoids the "unknown pragma" warnings from Clang-cl (Clang for Windows, b2 toolset=clang-win.)


robertramey:
slightly modified this to be easier to maintain.  checked it in and now testing.

PR #233, opened 2021-05-14
Remove useless b2.log file.

Romain-Geissler-1A:

PR #235, opened 2021-05-14
Remove useless autogenerated html files in test folder.

Romain-Geissler-1A:
These files seems to be autogenerated. I doubt you actually meant to commit it.


robertramey:
I DO mean to commit it.

The "normal" boost practice is to create documentation as a separate "job" which leaves the documentation in some other place. (The boost.org website? who knows.)  I want users of this library in either develop or master to have immediate access to the relevant documentation regardless of whether they are on line or not. That is, the boost serialization library attempts to be self contained.  My personal opinion is the the normal boost way doesn't not serve developers needs well.

Of course this doesn't prevent from boost using it's own method - as it does.

There's no conflict here.

Romain-Geissler-1A:
Ok, make sense, I am fine with that

However if you really do expect this doc to be available for everybody without being online, since people do use "master" branch when they use official Boost release, then these files shall be on the master branch too ;) (following the discussion we had already in https://github.com/boostorg/serialization/pull/236)

robertramey:
I do expect these docs to be on the master branch also.  Are they not there? In my world, after merging from develop to master, all files should be the same and in sync.

robertramey:
My mistake. The library.html and links.html are not meant to be uploaded.  The contain the results of my local testing and aren't relevant.  The other *.html files are maintained by hand ( an actual pain ) should be handled just like source files.

Romain-Geissler-1A:
So shall this pull request be re-opened, or you will cherry pick it directly in develop ?

robertramey:
OK - should be done now

Romain-Geissler-1A:
Ok.

FYI, in https://github.com/boostorg/serialization/pull/233 I also removed b2.log (the empty one, in the root of the project, you removed the non empty one, in the test folder), which most likely is useless too.

PR #245, opened 2021-12-31
Documentation maintenance

jngrad:
Run spellchecker on source code comment lines and HTML documentation. Fix rendering issues in code samples, such as missing template parameters due to `<class T>` being interpreted as an opening HTML tag, by escaping special characters.


robertramey:
accepted

PR #256, opened 2022-05-04
Use rehash(count) instead of rehash(bucket_count), to avoid excessive pre-allocation.

BrianWeed:
On windows, the number of buckets in an unordered_map never decreases: https://godbolt.org/z/9jrMv6Gah
When an archive is saved, the bucket_count is the MAX that it ever was in that session, no matter what count is.
When that archive is loaded, even though count may be small, the rehash would allocate way more memory than is necessary.

This change solves that by using count for the rehash, which will pre-allocate the appropriate number of buckets, on each platform: https://godbolt.org/z/6fjfWGh47


robertramey:
question: if one loads a previously existing archive with the old count type, will it get automatically converted to the new count type?

PR #270, opened 2022-10-18
Update github actions

sdarwin:
If you're interested here is a newer Github Actions config, copied from boost-ci. It should finish successfully. 

During the tests, Windows is running out of disk space. All the jobs take a long time.  It would be good to reduce the number of tests and then re-enable windows.


robertramey:
Rather than reducing the number of tests, I'd rather just disable CI for the serialization library.  Right not, in spite of efforts expended by many, the CI tests are just heating up the planet.

PR #280, opened 2023-02-22
Declare non-ODR-used copy constructor

ecatmur:
The compiler-generated copy constructor is deprecated since C++11 on classes with user-declared copy assignment operator.

This change allows clean compilation with the -Wdeprecated-copy/-Wdeprecated-copy-with-user-provided-ctor flag.

### DIFF
--- a/src/basic_oarchive.cpp
+++ b/src/basic_oarchive.cpp
@@ -68,13 +68,6 @@ class basic_oarchive_impl {
                 return false;
             return class_id < rhs.class_id;
         }
-        aobject & operator=(const aobject & rhs)
-        {
-            address = rhs.address;
-            class_id = rhs.class_id;
-            object_id = rhs.object_id;
-            return *this;
-        }
         aobject(
             const void *a,
             class_id_type class_id_,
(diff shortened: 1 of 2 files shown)


robertramey:
Did you actually test this?  Library fails to compile after marging this change: 

`boost::archive::xml::assign_impl<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&) in xml_grammar.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)`

ecatmur:
> Did you actually test this? Library fails to compile after marging this change:
> 
> `boost::archive::xml::assign_impl<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>> const&) in xml_grammar.o ld: symbol(s) not found for architecture x86_64 clang: error: linker command failed with exit code 1 (use -v to see invocation)`

Apparently not, sorry. Not sure how that got through. #297 for the correct fix.

PR #282, opened 2023-04-29
Simplify no wstreambuf support.

jzmaddock:
See https://github.com/boostorg/serialization/pull/224.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -12,6 +12,8 @@ project boost/serialization
       <conditional>@include-spirit
 ;
 
+import ../../config/checks/config : requires ;
+
 SPIRIT_ROOT = [ modules.peek : SPIRIT_ROOT ] ;
 rule include-spirit ( properties * )
 {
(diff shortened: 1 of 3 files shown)


jzmaddock:
@robertramey : Looks like the CI is stalled indefinitely waiting for a runner - that's because the ubuntu-18 image has been deprecated (for example) and available runners are stupidly scarce.  Shall I update the images used as part of this PR?  The catch is that older gcc versions will no longer be available.  They *are* available with the drone CI runners that the cpp alliance provides for us though...

robertramey:
Don’t worry about it. Boost I never really works for me anyway. I’ll except the PR, sync, my own machine, and run the tests locally. I’m sure this will be fine with a change this modest. Many thanks to you both for your help on these very Obscure issues. Robert Rameywww.rrsd.comOn Apr 30, 2023, at 1:15 AM, jzmaddock ***@***.***> wrote:﻿
@robertramey : Looks like the CI is stalled indefinitely waiting for a runner - that's because the ubuntu-18 image has been deprecated (for example) and available runners are stupidly scarce.  Shall I update the images used as part of this PR?  The catch is that older gcc versions will no longer be available.  They are available with the drone CI runners that the cpp alliance provides for us though...

jzmaddock:
@robertramey, I added the requirement to all of the tests which had linker errors when building the tests with -DBOOST_NO_STD_LOCALE.  Basically anything which used the xml_archives anywhere.

ArielSilver:
pull request created

PR #284, opened 2023-06-28
Use containers to supported deprecated OS with GHA

mborland:
Github deprecated support for a number of old Ubuntu distributions so use containers and the Ubuntu toolchain PPA to keep support for older compilers.


mborland:
@robertramey This is good for review. It has the Github Actions CI system working again, and a few minor fixes to that end. If you want you can go into the repo settings and delete the webhook for appveyor since that doesn't seem to work. That'll give you the green check mark next to PRs/commits moving forward.

PR #290, opened 2023-08-25
Fix predicate's defaulted copy constructor for C++03 while also fixing #276.

Romain-Geissler-1A:
Hi,

This fixes the issue I introduced in #276 (sorry for the breakage).

This time I tested not directly with boost, but using compiler explorer. This is the code I tested:

```
#include "boost/config.hpp"    

struct predicate {
    BOOST_DEFAULTED_FUNCTION(predicate(const predicate& rhs), : m_ti(rhs.m_ti) {})
    BOOST_DELETED_FUNCTION(predicate & operator=(const predicate & rhs))
public:
    const void * const m_ti;
    /*
    bool operator()(helper_value_type const &rhs) const {
        return m_ti == rhs.first;
    }
    */
    predicate(const void * ti) :
        m_ti(ti)
    {}  
};

void f(const predicate& a)
{
    predicate b(a);
    (void)b;
}
```

I tried with flags `-Wall -Wextra -Werror -std=gnu++03` and `-Wall -Wextra -Werror -std=gnu++23`, and with both gcc and clang x86_64 (trunk version in both cases). There is no compilation error in all these 4 different variants.

Compiler explorer link for clang/C++03: https://godbolt.org/z/PhPGsMdoM

Cheers,
Romain

### DIFF
--- a/include/boost/archive/detail/helper_collection.hpp
+++ b/include/boost/archive/detail/helper_collection.hpp
@@ -55,7 +55,7 @@ class helper_collection
     collection m_collection;
 
     struct predicate {
-        // BOOST_DEFAULTED_FUNCTION(predicate(const predicate& rhs) : m_ti(rhs.m_ti), {})
+        BOOST_DEFAULTED_FUNCTION(predicate(const predicate& rhs), : m_ti(rhs.m_ti) {})
         BOOST_DELETED_FUNCTION(predicate & operator=(const predicate & rhs))
     public:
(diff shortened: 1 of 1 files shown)


Romain-Geissler-1A:
Note: given that you use Boost filesystem in your tests (but not in the core library), and given that Boost filesystem 1.84 will drop C++03 support, you may hit some issues testing C++03 in the future.

robertramey:
merged and tested

PR #297, opened 2023-08-29
Implement copy constructor

ecatmur:
#280 take 2 (test matrix: toolset=clang-14,gcc-11 variant=debug,release cxxstd=11,14,17,20,2b)

The compiler-generated copy constructor is deprecated since C++11 on classes with user-declared copy assignment operator.

This change allows clean compilation with the -Wdeprecated-copy/-Wdeprecated-copy-with-user-provided-ctor flag.


robertramey:
merged and tested

PR #301, opened 2023-11-24
Make BOOST_STRONG_TYPEDEF constexpr

jakubmiernik:
Lack of constexpr constructor for BOOST_STRONG_TYPEDEF  cause errors in clang and warning in gcc when using in constexpr functions.


robertramey:
Suppose we make this change.  What happens when we use it with a compiler which does not support constant expressions?  I think this needs to be more elaborate to be totally portable.  I see merit in the idea, but I don't see it being portable to all compilers that the boost serialization library supports.

FYI. The serialization library no longer uses this macro.  It's been left in there as it seems that many find it useful.

PR #42, opened 2016-05-14
Fix undefined reference to codecvt_null<wchar_t>

xantares:
See https://svn.boost.org/trac/boost/ticket/12205

PR #302, opened 2023-11-29
fix "build/Jamfile.v2" - fix linking error when using wchar_t

gh-user-2022:
I can provide the exact linker error and/or reproducible example, if necessary.

(This is a very old bug, present in boost-1.69 (and possibly before), I had to patch the boost for several years.)

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -107,6 +107,7 @@ WSOURCES =
     xml_woarchive
     polymorphic_xml_wiarchive
     polymorphic_xml_woarchive
+    codecvt_null
 ;
 
 lib boost_serialization 


robertramey:
I have a question about this.

The b2 file contains: 
SOURCES =    ...
   codecvt_null

which means that this code is included in the library "boost_serialization".  Hence it should not be necessary in the library boost_wserialization. So it seems to me that making this change would mean the code is included twice.  Perhaps the code IS in boost_serialization but there might be some issue with visibility setting which prevents the linking to the code in boost_serialization.  Or ... I don't know.  Needs more investigation.

gh-user-2022:
There is following error without this patch:

```
[build] ld: ......./libboost_wserialization-gcc112-mt-d-1_69.a(basic_text_wiprimitive.o): in function `boost::archive::codecvt_null<wchar_t>::codecvt_null(unsigned long)':
[build] ......./boost/archive/codecvt_null.hpp:97: undefined reference to `vtable for boost::archive::codecvt_null<wchar_t>'
```

gh-user-2022:
And apparently there was already a patch for this problem, but it was rejected: https://github.com/boostorg/serialization/pull/42

Issue #319, opened 2024-08-11
std::optional serialization conflicts with boost::shared_ptr in Boost 1.84 and 1.85


sowle:
I'm experiencing a very strange compilation issue when using boost::shared_ptr with the (de)serialization of std::optional.

std::optional serialization support has been added recently in version 1.84.0: https://www.boost.org/doc/libs/1_84_0/boost/serialization/optional.hpp (see also commit [61a2b1207619cedb1f4d3ac1c7193f10ef9334f7](https://github.com/boostorg/serialization/commit/61a2b1207619cedb1f4d3ac1c7193f10ef9334f7))

After a lot of effort I have reduced the problem to a minimal reproducible example, which highlights the issue **in both 1.84 and 1.85** versions (earlier versions don't support std::optional serialization).

MRE:
```
#include <fstream>
#include <iostream>
#include <optional>
#include <boost/filesystem/fstream.hpp>
#include <boost/serialization/optional.hpp>
#include <boost/archive/text_iarchive.hpp>
#include <boost/archive/text_oarchive.hpp>

struct pluto_t
{
  uint64_t mass = 0;
  std::optional<uint64_t> optional_mass;

  template <class archive_t>
  void serialize(archive_t &_a, const unsigned int ver)
  {
    _a & mass;
    _a & optional_mass;                          // <-- no error if this line is commented out
  }

};


int main()
{
  const char* filename = "tmpfile";
  // store to file
  {
    pluto_t pluto{};
    pluto.mass = 13025000000000000000ull;

    std::ofstream ofs(filename);
    boost::archive::text_oarchive oa(ofs);
    oa << pluto;
  }

  // load from file
  {
    pluto_t pluto{};

    std::ifstream ifs(filename);
    boost::archive::text_iarchive ia(ifs);
    ia >> pluto;                                 // <-- no error if this line is commented out

    std::cout << pluto.mass << std::endl;
  }

  std::cout << "done." << std::endl;
  return 0;
}


#include <boost/serialization/shared_ptr.hpp>    // <-- no error if this line is commented out

[[maybe_unused]] static int foo = 42;
```

Error (there's lots of text, here's only the error line):
```
/usr/include/c++/8/type_traits:1147:12: error: invalid use of incomplete type ‘class boost::serialization::U’
     struct __is_trivially_copy_constructible_impl<_Tp, true>
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

[...]

PR #320, opened 2024-08-12
Alternative load declaration in shared_ptr_helper

correaa:
Attempt to solve https://github.com/boostorg/serialization/issues/319 , using the technique discussed in the thread


gennaroprota:
Thank you for the diagnosis. It is correct and is what the fix rests on.

I have not used this form, though, as declaring `load` for `SPT<U>` with any single-argument class template becomes ambiguous. I removed the declaration, instead, and the friendship for it.

Closing this in favor of PR #372.

PR #330, opened 2025-06-08
Move t instead of copying the value

venkat-murty:
`std::optional<T>` does not compile when T has a move operator and the copy operator is deleted.


gennaroprota:
Thanks. Modified a bit into PR #348, with a regression test.

PR #332, opened 2025-08-08
fix CVE-2023-34399

TrebledJ:
Negative indexing bug. Left the null tag commented for posterity/search purposes.

Reference: https://securelist.com/mercedes-benz-head-unit-security-research/115218/


gennaroprota:
Thanks. I expanded the message to describe the bug, rather than just name the CVE; the change itself is exactly yours.

PR #334, opened 2025-10-01
Fix required CMake version

Flamefire:
`cxx_std_11` requires CMake 3.8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the Boost Software License, Version 1.0.
 # https://www.boost.org/LICENSE_1_0.txt
 
-cmake_minimum_required(VERSION 3.5...3.16)
+cmake_minimum_required(VERSION 3.8...3.16)
 
 project(boost_serialization VERSION "${BOOST_SUPERPROJECT_VERSION}" LANGUAGES CXX)
 


pdimov:
There aren't any mentions of `cxx_std_11` in `CMakeLists.txt` though.

Flamefire:
IIRC I did it because some dependency does which "bubbled up" here and caused errors. Seen in failures of the [workflow testing CMake versions](https://github.com/boostorg/cmake/blob/develop/.github/workflows/cmake_versions.yml)

It doesn't matter much any more as we declared CMake 3.8 as the baseline for Boost: https://github.com/boostorg/boost/blob/7331414cbb8a1d645754a5229fd02f68da9d0103/CMakeLists.txt#L5

Flamefire:
@pdimov: Can you take a look at this:
https://github.com/boostorg/serialization/blob/097a6c63a137be836d663cdb27f2e6c803a4100b/CMakeLists.txt#L116-L118

It looks like this would do the installation twice on CMake >= 3.13

pdimov:
Why would it do the installation twice?

Flamefire:
The super-project calls `boost_install` for all libraries. Can this interfere with the manual call to `boost_install` here?

pdimov:
No, the manual call disables the automatic one. In fact I'm pretty sure you touched that logic; it used to look at the include directory, and then we changed it to use a dedicated property or something similar.

pdimov:
https://github.com/boostorg/cmake/commit/005e32d9318fa7bbeca501ae09c116ef5246d91a

And the manual call is required here, because the library has two targets to install, whereas the automatic one only installs a single target.

Flamefire:
Ah you are right, thanks for the remainder

PR #339, opened 2025-12-27
Allow variant deserialization to make types with private constructors

KJTsanaktsidis:
If a type has a private no-args constructor, but has friended boost::serialization::access, it can usually be deserialised as normal. However, it _can't_ be deserialized out of a variant, because the variant tries to invoke the default constructor.

Fix this by constructing it through boost::serialization::access, which is designed to model "classes that can't be created with no arguments except through deserialization"

### DIFF
--- a/include/boost/serialization/access.hpp
+++ b/include/boost/serialization/access.hpp
@@ -129,6 +129,10 @@ public:
         // class doesn't have a class-specific placement new defined.
         ::new(t)T;
     }
+    template<class T>
+    static T *construct_r(void * t) {
+        return ::new(t)T;
+    }
     template<class T, class U>
     static T & cast_reference(U & u){
         return static_cast<T &>(u);
--- a/include/boost/serialization/std_variant.hpp
+++ b/include/boost/serialization/std_variant.hpp
@@ -122,11 +122,19 @@ struct variant_impl
             // with an implementation that de-serialized to the address of the
             // aligned storage included in the variant.
             using type = mp::front<Seq>;
-            type value;
-            ar >> BOOST_SERIALIZATION_NVP(value);
-            v = std::move(value);
+            struct ValueData {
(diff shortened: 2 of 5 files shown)


include/boost/serialization/access.hpp
```diff
@@ -129,6 +129,10 @@ class access {
         // class doesn't have a class-specific placement new defined.
         ::new(t)T;
     }
+    template<class T>
+    static T *construct_r(void * t) {
```
KJTsanaktsidis, line 133: Hm, had to add a new overload of `construct` that actually returned the placement-new result, because of pointer provenance issues.

```
alignas(T) unsigned char mem[sizeof(T)];
T *ptr = reinterpret_cast<T*>(mem);
::new(ptr)T;

ptr->stuff();
```

Apparently is no bueno because the compiler can assume that ptr is NOT the same thing as the in-place-constructed T. Need to instead do something like

```
alignas(T) unsigned char mem[sizeof(T)];
T *ptr = ::new(mem)T;

ptr->stuff();
```

gennaroprota:
Your diagnosis is right, and thank you for the PR. Two things had to be fixed, though.

The header your patch changes is not the one in play. Since the 2023 re-import, variant.hpp handles all three flavors, `boost::variant`, `variant2` and `std::variant`, and it carries the same `head_type value;` at line 162. Nothing in the library, tests or examples includes std_variant.hpp, and including it alongside variant.hpp is a redefinition error, so a fix there would not reach your own case if you include variant.hpp. I'll address the orphaned headers issue separately.

The library also already has the machinery you hand-rolled. `detail::stack_construct` is aligned storage plus `load_construct_data_adl` plus destruction, and it is what optional.hpp uses for exactly this problem since [#121](https://github.com/boostorg/boost/issues/121), so `access::construct_r` is not needed.

PR #340, opened 2026-01-22
Remove dependencies on Boost.StaticAssert

Lastique:
Boost.StaticAssert has been merged into Boost.Config, so remove the dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,6 @@ target_link_libraries(boost_serialization
     Boost::preprocessor
     Boost::smart_ptr
     Boost::spirit
-    Boost::static_assert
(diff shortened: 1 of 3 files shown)


pdimov:
I'm going to merge this because it's necessary. Without it, we get errors like
```
error: Unable to find file or target named
error:     '/boost/static_assert//boost_static_assert'
error: referred to from project at
error:     '/D:/a/boost_install/boost-root/libs/serialization/build'
error: could not resolve project reference '/boost/static_assert'
```
because `boost/static_assert.hpp` has been moved to Config, so its inclusion no longer introduces a dependency on StaticAssert and doesn't cause it to be installed.

PR #342, opened 2026-05-05
Enable adding Ubuntu toolchain PPA by default

Flamefire:
The default in Boost.CI will be false but some jobs still rely on this, so keep it for now the default.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -8,7 +8,7 @@
 # As the yaml syntax for Drone CI is rather limited.
 #
 #
-globalenv={}
+globalenv={'ADD_UBUNTU_TOOLCHAIN_PPA': 'true'}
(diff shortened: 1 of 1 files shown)


gennaroprota:
Thank you for the patch and for maintaining Boost.CI.

Before I merge it, I wonder whether we are still using Drone:

- No Drone checks appear on recent pull requests.
- drone.star has not changed since February 2021.

If we are using it, I'll merge your PR. If we are not, I think it's better to delete drone.star completely.

Flamefire:
It seems like you are not using Drone: https://drone.cpp.al/boostorg/serialization/settings
> This Repository is not Active.

So I guess you can just delete your Drone config.

I'll close this PR as any changes to Boost.CI for Drone won't affect you then.

gennaroprota:
Thanks.

PR #358, opened 2026-07-29
Give collection_size_type explicit increment, decrement, and stream extractor

### LINKED ISSUE #192: collection_size_type missing increment/decrement operators
Various places such as [here](https://github.com/boostorg/serialization/blob/develop/include/boost/serialization/vector.hpp#L191) use `operator--` or `operator++` of `collection_size_type`, but [these are not defined](https://github.com/boostorg/serialization/blob/develop/include/boost/serialization/collection_size_type.hpp#L20).

This doesn't seem to be a problem for most compilers, but it's apparently causing [compile errors with Fedora 32](https://github.com/freeorion/freeorion/issues/2795) for my project:
````
/usr/include/boost/serialization/vector.hpp: In function ‘void boost::serialization::save(Archive&, const std::vector<bool, Allocator>&, unsigned int)’:
/usr/include/boost/serialization/vector.hpp:191:16: error: no post-decrement operator for type
  191 |     while(count-- > 0){
      |                ^~
````

gennaroprota:
`collection_size_type` had no increment, decrement, or stream extractor but had an implicit conversion to a modifiable reference to its stored size (`operator base_type &`) which allowed the built-in operators to apply.

Remove that mutable-reference conversion and provide the operations the library actually needs explicitly: pre/post `++` and `--`, and an `operator>>` for input (templated on the character type, so it serves narrow and wide text and XML archives). The read-only `operator base_type() const` is kept: output, container sizing, and comparisons go through it, so no call sites change.

The on-disk format is unchanged. Binary archives already serialize the type bitwise; text and XML still read and write the same size through `is >> x` / `os << x`.

Adds test_collection_size_type, which checks the operators directly and round-trips a value through every archive type.

Closes #192.

### DIFF
--- a/include/boost/serialization/collection_size_type.hpp
+++ b/include/boost/serialization/collection_size_type.hpp
@@ -2,11 +2,13 @@
 #define BOOST_SERIALIZATION_COLLECTION_SIZE_TYPE_HPP
 
 // (C) Copyright 2005 Matthias Troyer
+// Copyright 2026 Gennaro Prota
 // Distributed under the Boost Software License, Version 1.0.
 // (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
 #include <cstddef> // size_t
+#include <istream>
 #include <boost/serialization/strong_typedef.hpp>
 #include <boost/serialization/level.hpp>
 #include <boost/serialization/split_free.hpp>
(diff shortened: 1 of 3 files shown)


gennaroprota:
Closing this without merging. See https://github.com/boostorg/serialization/issues/192#issuecomment-5127662236.

